PR #31, opened 2021-01-09
Add warning about changing import code on revisions

CactusPuppy:
It is often the case that newer Workshoppers will not realize that they are able to upload updates to their gamemodes to the existing associated import code. This warning will help educate those people.


Mitcheljager:
Looks great, thank you! Merging it in

PR #32, opened 2021-01-09
Add Korean translation for PTR warning label

CactusPuppy:


Mitcheljager:
Ah yes, forgot about that one. Thank you!

PR #35, opened 2021-02-04
Remedy invalid reports

CactusPuppy:
TODO: 
- Handle clicking on an existing invalid report (destroy/reject?)
- Make a Rake task to clean up invalid reports


app/views/admin/reports/show.html.erb
```diff
@@ -13,17 +13,30 @@
...
+  <% if @report.reported_user.present? %>
+    <div class="mb-1/2">
+      <strong>Reported Content Author</strong>:<br>
+      <%= @report.user.username %>
+      <small>(<%= @report.user.id %>)</small> -
+      <em><%= link_to "view user", admin_user_path(@report.user_id) %></em>
     </div>
   <% end %>
 
+  <%# TODO: Refactor into partial %>
```
CactusPuppy, line 25: Still need to refactor this to future-proof for the addition of new types of reports

CactusPuppy (commented):
Just need to refactor partial in the report show page

app/controllers/admin/reports_controller.rb
```diff
@@ -5,6 +5,14 @@ def index
 
   def show
     @report = Report.find(params[:id])
-    @post = Post.find(@report.concerns_id) if @report.concerns_model == "post"
+    begin
```
Mitcheljager, line 8: I think it'd be better to simply check `@post.present?` instead.

app/views/admin/reports/show.html.erb
```diff
@@ -13,17 +13,17 @@
...
+  <% if @report.reported_user.present? %>
+    <div class="mb-1/2">
+      <strong>Reported Content Author</strong>:<br>
+      <%= @report.user.username %>
+      <small>(<%= @report.user.id %>)</small> -
+      <em><%= link_to "view user", admin_user_path(@report.user_id) %></em>
     </div>
   <% end %>
 
+  <%= render "#{ @report.concerns_model }", report: @report, post: @post %>
```
Mitcheljager, line 25: The string interpolation here is unnecessary

app/helpers/reports_helper.rb
```diff
@@ -10,4 +10,21 @@ def report_options
...
+    if report_has_model_property?(report, model, property)
+      return report.properties[model][property]
+    else
+      return ""
+    end
+  end
+
+  private
+  def report_has_model_property?(report, model, property)
+    return false unless report.respond_to? :properties
```
Mitcheljager, line 24: I've never seen `.respond_to?` before, but won't this always be true?

CactusPuppy: [Documentation on respond_to](https://apidock.com/rails/ActionController/MimeResponds/InstanceMethods/respond_to)

This check is to handle old reports whose properties field will be `nil` since it was created without any properties. Nil does not respond to the properties method call, and thus the check will return early.

app/controllers/admin/reports_controller.rb
CactusPuppy: The issue is that `Post.find` doesn't return `nil` if the record is not found, it instead raises an `ActiveRecord::RecordNotFound` exception. I suppose using `Post.where.take` would prevent that issue, however.

app/helpers/reports_helper.rb
CactusPuppy: If the `respond_to?` check wasn't there, the method would error when trying to process old reports.

app/controllers/admin/reports_controller.rb
Mitcheljager: Perhaps using `find_by` will also do the trick, I'm not sure.

db/schema.rb
```diff
@@ -223,6 +223,8 @@
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
     t.string "visit_token"
+    t.text "properties"
```
Mitcheljager, line 226: Another option might also be to add `, default: "[]"` here to prevent old reports from failing without the multiple checks. I'm totally fine with your method as well though.

CactusPuppy: I think I'll go ahead and add that to keep the helper methods cleaner.

app/controllers/admin/reports_controller.rb
CactusPuppy: `find_by` would probably work better yeah

app/helpers/reports_helper.rb
```diff
@@ -10,4 +10,19 @@ def report_options
...
+      return report.properties[model][property]
+    else
+      return ""
+    end
+  end
+
+  private
+  def report_has_model_property?(report, model, property)
+    return false unless report.properties[model]
+    return false unless report.properties[model].respond_to?(:[])
```
CactusPuppy, line 25: Technically this line isn't necessary since old reports will fail the first check, but I am keeping it in just in case future additions store non-accessible content to `properties[model]`.

Mitcheljager (approved):

PR #36, opened 2021-02-05
Improve text and warnings on report form

ZusorCode:
Add new options to report form:
`There's a bug in this code` and `I can't import this code`
Selecting them reveals additional warnings when reports should not be submitted.
This should reduce the amount of `Other` reports reporting a bug in a code.
image
image

### DIFF
--- a/app/views/reports/_form.html.erb
+++ b/app/views/reports/_form.html.erb
@@ -22,8 +22,8 @@
         Do not report if:
 
         <ul class="mt-1/8 pl-1/2">
-          <li>This code is less than 6 months old and has not expired.</li>
-          <li>This code not for PTR.</li>
+          <li>This code has expired.</li>
+          <li>This code is made for PTR.</li>
         <ul>
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


app/views/reports/_form.html.erb
```diff
@@ -10,6 +10,24 @@
...
+      </div>
+    </div>
+
+    <div data-reveal-by-select-target="I can't import this code" class="visibility-hidden">
+      <div class="block mt-1/4 text-small text-dark">
+        Do not report if:
+
+        <ul class="mt-1/8 pl-1/2">
+          <li>This code is less than 6 months old and has not expired.</li>
+          <li>This code not for PTR.</li>
```
Mitcheljager, line 26: This sentence is a little awkward

CactusPuppy: I'd suggest "This code is made for PTR" for the second sentence

ZusorCode: Would this work better?
Do not report if:
- This code has expired
- This code is made for PTR

Mitcheljager: Yes, perfect :)

Mitcheljager:
Merging it in, thank you :)

PR #39, opened 2021-02-17
List destroyed comments in the admin activity log

CactusPuppy:


Mitcheljager:
I'm a bit confused, isn't this exactly how it is already? https://github.com/EloHellEsports/workshop.codes/blob/6de0f89ee63e88cf2aab7f03678873967fdba2c7/app/views/admin/base/index.html.erb#L83

CactusPuppy:
Huh, could've sworn it wasn't there. No idea how I didn't catch that.

PR #40, opened 2021-02-25
Don't leave report page if report is accepted

CactusPuppy:
Usually after a report is accepted, admins will want to take action on the reported content.


Mitcheljager (approved):
Sounds like a good idea!

PR #42, opened 2021-04-08
Prevent users from submitting reports under "There's a bug in this code" category

CactusPuppy:
This simply enforces the stipulation in the report form that users should not submit reports under this category.


Mitcheljager (approved):
Nice one!

PR #44, opened 2021-05-07
Add supported player counts

CactusPuppy:
This pull request adds support for filtering posts by supported player counts. 

Example of post listings with player counts
image

Example of post settings page
image

Search filter view
image


app/views/posts/_post.html.erb
```diff
@@ -29,6 +29,12 @@
           <% end %>
         </div>
 
+        <div class="item__details-item">
```
Mitcheljager, line 32: I don't think I like this to be honest, I think I'd rather keep the post overview as clean as possible.

app/views/filter/_filter.html.erb
```diff
@@ -96,6 +96,15 @@
           </div>
         </div>
 
+        <div class="filter__group mt-1/4 mb-1/1">
+          <p class="filter__label">Players</p>
+          <%= content_tag :div, nil, class: "num-player-slider", data: {
```
Mitcheljager, line 101: I've personally been using `tag.div` rather than `content_tag`, I think for the sake of consistency this one should be changed.

app/views/filter/_filter.html.erb
```diff
@@ -96,6 +96,15 @@
...
+            type: "filter",
+            min_players: (to_range(CGI.escape(params[:players])).first if params[:players].present?),
+            max_players: (to_range(CGI.escape(params[:players])).last if params[:players].present?) } %>
+          <div style="display: none" data-filter-type="players" <%= "data-value=#{CGI.escape(params[:players])}" if params[:players].present? %>></div>
```
Mitcheljager, line 105: This could also be a rails tag rather than a html tag. Also, there's a `hidden` class you could use instead of `style="..."`, not strictly necessary, but a little nicer in my opinion.

app/assets/javascripts/application.js
```diff
@@ -14,6 +14,7 @@
 //= require turbolinks
 //= require microlight
 //= require timeago.js/dist/timeago.min
+//= require nouislider/distribute/nouislider.min
```
Mitcheljager, line 17: I think it would make sense to require this in `num-players-slider.js` rather than in this file.

app/assets/javascripts/num-players-slider.js
```diff
@@ -0,0 +1,73 @@
...
+document.addEventListener("turbolink:before-cache", function() {
+  if (sliders.length == 0) return;
+
+  sliders.forEach(slider => { slider.noUiSlider.destroy() });
+
+  sliders = []
+})
+
+document.addEventListener("turbolinks:load", function() {
+  const elements = document.querySelectorAll(".num-player-slider");
```
Mitcheljager, line 12: For consistency please use a data attribute like `[data-role='num-player-slider']` (or `[data-num-player-slider]`). That way classes are strictly used for styling and data-attributes strictly used for javascript.

app/assets/javascripts/num-players-slider.js
```diff
@@ -0,0 +1,73 @@
...
+        'max': 12
+      },
+      step: 1,
+      pips: {
+        mode: 'steps'
+      },
+      behaviour: 'tap-drag'
+    });
+
+    switch (element.dataset.type) {
```
Mitcheljager, line 39: I'm not a big fan of switch cases, especially when it's only 2 cases. 
A nicer solution might be:
```
if (element.dataset.type == "post") element.noUiSlider.on("set", postOnSliderUpdate)
if (element.dataset.type == "filter") element.noUiSlider.on("set", filterOnSliderUpdate)
```

app/assets/stylesheets/form/_slider.scss
```diff
@@ -0,0 +1,298 @@
+/*! nouislider - 14.7.0 - 4/6/2021 */
+/* Functional styling;
+ * These styles are required for noUiSlider to function.
+ * You don't need to change these rules to apply your design.
+ */
+ .noUi-target,
+ .noUi-target * {
+   -webkit-touch-callout: none;
```
Mitcheljager, line 8: I know these are just copy pasted from their thingie, but let's remove browser prefixes, those get added automatically anyway (and most are incredibly outdated)

app/controllers/filter_controller.rb
```diff
@@ -15,6 +15,7 @@ def index
     @posts = @posts.select { |post| to_slug(post.categories).include?(to_slug(params[:category])) } if params[:category]
     @posts = @posts.select { |post| to_slug(post.maps).include?(to_slug(params[:map])) } if params[:map]
     @posts = @posts.select { |post| to_slug(post.heroes).include?(to_slug(params[:hero])) } if params[:hero]
+    @posts = @posts.select { |post| to_range(params[:players]).overlaps?((post.min_players)..(post.max_players)) } if params[:players]
```
Mitcheljager, line 18: Never seen this before, neat!

db/migrate/20210427162657_add_player_counts_to_posts.rb
```diff
@@ -0,0 +1,6 @@
+class AddPlayerCountsToPosts < ActiveRecord::Migration[6.1]
+  def change
+    add_column :posts, :min_players, :integer, default: 1
```
Mitcheljager, line 3: Just adding this as a reminder that this should default to `null` or whatever so old posts don't show up in every result even though they might not be relevant

app/assets/javascripts/application.js
CactusPuppy: Makes total sense. Didn't know you could include within files beyond the `application.js`.

app/assets/javascripts/num-players-slider.js
CactusPuppy: I made it a switch case so that if more cases needed to be added, I could do so easily. However, since it doesn't seem like more than two cases are needed, I agree the switch statement is overkill.

app/views/posts/_post.html.erb
CactusPuppy: Fair enough. I'll let you determine where you want to display the players.

Mitcheljager (approved):

PR #53, opened 2021-06-08
Derivatives (minus public displays)

CactusPuppy:
# Overview

This pull request aims to establish a system by which users can specify which post or posts their post builds on. While it is possible to credit other posts within the description of a post, this system will add a universal crediting section which will be more consistent for visitors.


.ruby-version
```diff
@@ -1 +1 @@
-ruby-3.0.0
+ruby-3.0.1
```
CactusPuppy, line 1: Bumping the Ruby version was necessary to fix some tests incorrectly erroring locally.

Gemfile
```diff
@@ -47,11 +47,11 @@ group :development, :test do
...
   gem "capybara", "~> 2.13"
   gem "database_cleaner"
-  gem "factory_bot_rails", "~> 4.0"
+  gem "factory_bot_rails", "~> 6.2"
   gem "faker"
-  gem "rspec-rails", "~> 4.0.1"
+  gem "rspec-rails", "~> 4.0.2"
   gem "selenium-webdriver"
-  gem "shoulda-matchers", "~> 3.1"
+  gem "shoulda-matchers", "~> 4.5"
```
CactusPuppy, line 54: Similar to `.ruby-version`, these gems were bumped to address local test failures.

app/assets/stylesheets/general/_form.scss
```diff
@@ -239,6 +239,10 @@ input[type="file"] {
   label + & {
     margin-top: $margin / 3 - $margin / 6;
   }
+
+  &--left {
```
Mitcheljager, line 243: There also is a `text-left` utility class that does the same

app/assets/stylesheets/general/_form.scss
```diff
@@ -312,3 +316,75 @@ input[type="file"] {
   font-size: $font-size-small;
   color: $text-color-dark;
 }
+
+.form-tags-wrapper {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  padding: 0 5px 5px 5px;
```
Mitcheljager, line 324: This is more of a personal preference thing, but I've been avoiding using pixels in favor of rem. There's 2 advantages;
- Flexibility; If we ever decide to change the base font size everything else will scale accordingly.
- Consistency. A rem value is easier to relate to something else, given you use nice round values. For example it's easier to understand why you might use .5rem than it is why you might use 8px. 

In this case of course 5px is nice and round, but other values are not. I'd suggest going with .25rem here.

If you disagree please let me know!

app/assets/stylesheets/general/_form.scss
```diff
@@ -312,3 +316,75 @@ input[type="file"] {
...
+  padding: 4px 5px;
+}
+
+.form-tags-tag {
+  display: flex;
+  white-space: nowrap;
+  border-radius: 2px;
+  margin-right: 5px;
+  margin-top: 5px;
+  padding: 4px 2px 3px 4px;
```
Mitcheljager, line 340: This is a better example for my previous comment. `4 2 3 4`, Why were those values chosen? Something like `.25rem, .15rem, .2rem, .25rem` makes more sense in my mind, but even then you could probably go with just `.25rem, .15rem` without losing much value (I could be wrong here since I'm not sure what it looks like).

app/assets/stylesheets/general/_form.scss
```diff
@@ -312,3 +316,75 @@ input[type="file"] {
...
+  left: 0;
+  right: 0;
+  margin: 3px 0;
+  padding: 5px;
+  background: $bg-dark;
+  border: solid 1px $border-color;
+  border-radius: 5px;
+  max-height: 310px;
+  overflow: auto;
+  overflow-x: hidden;
```
Mitcheljager, line 368: Double overflow declaration here, `overflow-y: auto` alone should do the job

app/assets/stylesheets/general/_form.scss
```diff
@@ -312,3 +316,75 @@ input[type="file"] {
   font-size: $font-size-small;
   color: $text-color-dark;
 }
+
+.form-tags-wrapper {
```
Mitcheljager, line 320: Would it make more sense to move all `form-tags` classes to their own scss file?

app/assets/stylesheets/general/_form.scss
```diff
@@ -312,3 +316,75 @@ input[type="file"] {
...
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  padding: 0 5px 5px 5px;
+}
+
+.form-tags-input {
```
Mitcheljager, line 327: These class names don't necessarily follow BEM. The wrapper should be called something like `.form-tags` and this child would be `.form-tags__input`, and similarly for other classes used here.

I don't super mind, so feel free this ignore this one if you don't feel like changing it.

app/assets/stylesheets/general/_form.scss
```diff
@@ -312,3 +316,75 @@ input[type="file"] {
...
+  background: $bg-dark;
+  border: solid 1px $border-color;
+  border-radius: 5px;
+  max-height: 310px;
+  overflow: auto;
+  overflow-x: hidden;
+  z-index: 100;
+  text-align: left;
+
+  li {
```
Mitcheljager, line 372: Would be nice if we can give this a class instead of using the `li` tag.

app/javascript/packs/logged-in-user.js
```diff
@@ -40,3 +42,8 @@ document.addEventListener("turbolinks:before-cache", function() {
   const svelteComponents = document.querySelectorAll("[data-svelte-component]")
   svelteComponents.forEach(element => element.innerHTML = null)
 })
+
+document.addEventListener("turbolinks:click", () => {
```
Mitcheljager, line 46: I'm a bit confused why this is here, isn't this the same as the previous line, but on `click` instead of `before-cache`?

app/javascript/src/components/form/Derivatives.svelte
```diff
@@ -0,0 +1,84 @@
+<script>
+  import Tags from './Tags.svelte';
```
Mitcheljager, line 2: I see the js-linter isn't work in Svelte files. It's fine for now, I will see if I can get it fixed.

app/javascript/src/components/form/Derivatives.svelte
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  async function handleAutoCompleteRequest(value) {
+    if (!value) return [];
+
+    const timeout = 8000;
+    const timeoutController = new AbortController();
+    const timeoutID = setTimeout(() => timeoutController.abort(), timeout);
+
+    const response = await fetch(`/code/${value}`, {
```
Mitcheljager, line 20: There is a `fetch-rails.js` import you could use for fetch requests instead. Could be used like

```
new FetchRails(`/code/${value}`).get()
  .then(data => {
      const json = JSON.parse(text)
      return json.map(post => postToResult(post))
  })
  .error(error => {
    throw new Error(`${response.status} ${response.statusText}`)
  })
``

Although you wouldn't have the timeout you created above, but perhaps it would make sense to add the timeout to the `fetch-rails.js` so it applies to everything.
I don't mind your solution here either, it seems nice enough. So feel free to decide if you want to do this or not.

File: app/javascript/src/components/form/Derivatives.svelte [hunk 13]
```diff
@@ -0,0 +1,84 @@
...
+      signal: timeoutController.signal
+    });
+    const text = await response.text();
+    clearTimeout(timeoutID);
+
+    if (response.ok) {
+      const json = JSON.parse(text);
+      return json.map(post => postToResult(post));
+    } else {
+      throw new Error(`${response.status} ${response.statusText}`);
```
Mitcheljager, line 35: We might need to actually handle the error. Just setting a variable and displaying it is probably enough

app/javascript/src/components/form/Derivatives.svelte
```diff
@@ -0,0 +1,84 @@
+<script>
+  import Tags from './Tags.svelte';
+  import { onMount } from 'svelte';
+
+  let showDerivative = false;
+  let maxCodes = 5;
+  export let currSources = [];
```
Mitcheljager, line 7: Small thing, but I try to stay away from abbreviations, especially for props. It's easier to remember a word than it is an abbreviation, in my opinion.

app/javascript/src/components/form/Derivatives.svelte
```diff
@@ -0,0 +1,84 @@
...
+    >
+    <label
+      class="switch-checkbox__label"
+      for="show_derivative"
+    >
+      This code is a fork/uses other codes
+    </label>
+  </div>
+
+  <div class="form-group mt-1/4" hidden={!showDerivative}>
```
Mitcheljager, line 64: You could wrap this block in { #if showDerivative } ... { /if } instead of setting it to hidden. That way nothing gets render for those that never enable it in the first place.

app/javascript/src/components/form/Derivatives.svelte
CactusPuppy: The error is caught in the Tags form via the `await` block.

See https://github.com/EloHellEsports/workshop.codes/pull/53/files#diff-cbf984e6496b634b0209250da97c2279b04e079aa8ec2330e0d0e307733c9b4dR205-R213
image

Mitcheljager: Ah! Perfect

app/javascript/packs/logged-in-user.js
CactusPuppy: For some reason I found that Turbolinks was still showing me a cached version of the page with Svelte components and this addition happened to resolve the issue. Unsure if this applies to a Heroku deployment. I may deploy to my own staging environment to have a look.

app/assets/stylesheets/general/_form.scss
CactusPuppy: It probably makes sense at this point to be fair.

app/javascript/src/components/form/Derivatives.svelte
CactusPuppy: I was unaware that `fetch-rails.js` existed. I'll probably add a timeout feature to it and DRY the code.

app/assets/stylesheets/general/_form.scss
CactusPuppy: I'm not entirely sure why, but the 1 pixel helps center things better. For the sake of not making things look arbitrary though, `.25rem, .15rem` doesn't lose much aesthetic value.

app/models/derivative.rb
```diff
@@ -0,0 +1,17 @@
...
+  validates :derivation_id, presence: true
+  validate :no_self_derive
+  validates :source, uniqueness: { scope: :derivation, message: 'already derived from for this post.' }, if: -> { source.present? }
+  validates :source_code, uniqueness: { scope: :derivation, message: 'already derived from for this post.' }
+
+  private
+
+  def no_self_derive
+    errors.add(:base, 'Post cannot derive from itself') if source == derivation
+    errors.add(:base, 'Cannot derive from the same import code') if source_code == derivation&.code
```
CactusPuppy, line 15: TODO: Make import code checks case-insensitive

PR #54, opened 2021-06-08
GitHub Actions tweaks

CactusPuppy:
This pull request changes two things about the current GH Actions flow:

- Linters are now run in parallel to specs
- Specs and linters will now run to completion regardless of fails/errors

The first change was made to facilitate faster CI runs. The second change was made to allow developers to see all the currently failing CI tests in a single run.﻿


Mitcheljager:
Makes sense, thank you!

PR #60, opened 2021-07-25
Show derivatives in post show

Mitcheljager:
**Draft!**

- Re-enable derivatives in form
- Show codes that this post derives from (sources) after post description
- Show posts that derive from this post in extra tab
- Add ESLint for Svelte
- Fix ESLint errors

The derivatives are shown at the end of the description. There's also a link included under the title which scrolls you down to the derivatives. Derivatives that are on Workshop.codes link to them, those that are not are static.

image

image


app/assets/stylesheets/elements/_derivatives.scss
```diff
@@ -0,0 +1,43 @@
...
+  font-size: 1.25rem;
+  font-weight: normal;
+  line-height: 1em;
+  font-family: "Barlow Condensed", "Impact";
+  text-transform: uppercase;
+}
+
+.derivatives-marker {
+  
+}
```
CactusPuppy, line 43: Shouldn't SCSS files not contain empty rulesets?

.eslintrc.yml
```diff
@@ -5,6 +5,13 @@ parserOptions:
   ecmaVersion: 12
   sourceType: module
   allowImportExportEverywhere: true
+plugins:
+  - svelte3
+overrides:
+  -
```
CactusPuppy, line 11: This line looks funny. Shouldn't there be a name beyond the hyphen? I'm not too familiar with the YAML syntax in this scenario.

app/views/derivatives/_derived_from.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="items-heading mt-0 mb-0">
+  <h2 class="mt-0 mb-0">Codes derived from this code</h2>
```
CactusPuppy, line 2: Again, I think we could get away with saying "Posts derived from this post"

app/views/derivatives/_derived_from.html.erb
```diff
@@ -0,0 +1,9 @@
...
+</div>
+
+<blockquote>
+  These codes use <strong><%= @post.code %></strong> as part of their code. This could be because they built off of it, used parts of it, used it for inspiration, or other reasons.
```
CactusPuppy, line 6: Same down here. I think "These posts use CODE as part of their code. etc. etc. etc." sounds much better.

CactusPuppy: Also, a bit of wordsmithing, but I think
```
This could be because they built off of it, used parts of it, used it for inspiration, or other reasons.
```
could be changed to
```
This could be because they remixed it, used parts of it, were inspired by it, or other reasons.
```

app/javascript/src/components/form/Derivatives.svelte
```diff
@@ -1,76 +1,82 @@
...
-        fetchAutoCompleteValues={handleAutoCompleteRequest}
-      />
+        fillValues={ currentSources }
+        hidden={ !showDerivative }
+        allowSpace={ false }
+        onlyAlphanumeric={ true }
+        onlyCaps={ true }
+        tagLimit={ maxCodes }
+        useAutoComplete={ true }
+        fetchAutoCompleteValues={handleAutoCompleteRequest } />
```
CactusPuppy, line 79: I think a space got missed here. Shouldn't this be
```
fetchAutoCompleteValues={ handleAutoCompleteRequest } />
```
More importantly, did the linter not catch this in Actions?

app/javascript/src/components/form/Tags.svelte
```diff
@@ -186,33 +195,34 @@
...
               { @html result.html }
             </li>
-          {/each}
-        {:else}
-          <li class="list-element"><span class="m-0">No results on Workshop.codes</span></li>
-        {/if}
-      {:catch error}
+          { /each }
+        { :else }
+          <li class="list-element"><em>No results on Workshop.codes, but you can still add the code.</em></li>
```
CactusPuppy, line 213: Good catch!

app/views/derivatives/_derivatives.html.erb
```diff
@@ -0,0 +1,21 @@
+<hr class="small" id="sources" />
+
+<h3 class="mb-1/4">Sources</h2>
+
+<p><em><small>This code uses parts of or entire other codes. This could be because they built off of them, used parts of them, used them for inspiration, or other reasons.</small></em></p>
```
CactusPuppy, line 5: I think "This post utilizes other codes either in part or in full." reads better. Also see below for some general wordsmithing.

CactusPuppy (changes requested):
I'm hyped! I think there's a few things that could be addressed, but the overall design is great. Thank you again for all your hard work!

.eslintrc.yml
Mitcheljager: Yeah it's weird, but this creates an object which is what the plugin needed. I tried a couple of other ways of writing it, non of it would do the job.

app/javascript/src/components/form/Derivatives.svelte
Mitcheljager: Unfortunately the linter only checks the <script> blocks.

CactusPuppy (approved):
LGTM!

PR #68, opened 2021-08-29
💄 Adjust SELECTION to the look and feel

byvide:


app/assets/stylesheets/general/_general.scss
```diff
@@ -123,3 +123,8 @@ strong {
   width: 100%;
   height: auto;
 }
+
+::selection {
+  color:white;
+  background-color: var(--primary);
```
Mitcheljager, line 129: I don't think it's wise to use the primary color, as the color might be lost. For example `<mark>` is makes the text colored primary, and buttons have it as their background.

Mitcheljager (changes requested):

app/assets/stylesheets/general/_general.scss
byvide: Well, the color is set to white so i dont see the conflict with highlighted texts, on the other hand despite only having 3 buttons with primary colored background, it s an issue, yes.

byvide (commented):
im a noob and learning github so dont look at this commit, its bad

Issue #71, opened 2021-09-04
Controls may sometimes be saved as `undefined`


CactusPuppy:
Under unknown circumstances, it is possible for a code to be saved with `undefined` as the controls instead of the usual `{}`. More investigation required.

PR #73, opened 2021-09-08
Fix controls + enable Svelte debugging

CactusPuppy:
This pull request accomplishes two minor goals:
- Enable loading of Svelte in dev mode within development environments. This change can be used in tandem with a tool such as [Svelte DevTools](https://github.com/RedHatter/svelte-devtools) to enable easier Svelte component debugging.
- Solves #71 by ensuring that undefined controls are re-saved as blank. However, if filled controls are somehow saved as `undefined`, a new issue should be opened.

### DIFF
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -182,7 +182,7 @@ class Post < ApplicationRecord
   end
 
   def parsed_controls
-    JSON.parse(self.controls)
+    JSON.parse(self.controls).presence || []
   end
 
   def self.find_by_code(code)
--- a/app/models/post.rb
+++ b/app/models/post.rb
@@ -185,7 +185,8 @@ class Post < ApplicationRecord
     begin
       JSON.parse(self.controls).presence || []
     rescue JSON::ParserError
-      self.update controls: "[]"
+      self.controls = "[]"
+      self.save touch: false
       retry
     end
   end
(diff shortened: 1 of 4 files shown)


app/models/post.rb
```diff
@@ -182,7 +182,12 @@ def expired?
   end
 
   def parsed_controls
-    JSON.parse(self.controls)
+    begin
+      JSON.parse(self.controls).presence || []
+    rescue JSON::ParserError
+      self.update controls: "[]"
```
Mitcheljager, line 188: This seems a little dangerous, but at the same time I kind of like it. I do feel like it needs `.save(touch: false)` so as to not update the `updated_at` column.

config/webpack/loaders/svelte.js
```diff
@@ -9,6 +9,9 @@ module.exports = {
       preprocess: {
         style: sass()
       },
+      compilerOptions: {
+        dev: process.env.NODE_ENV == 'development' ? true : false,
```
Mitcheljager, line 13: Trailing comma!

CactusPuppy: r/justa

app/models/post.rb
CactusPuppy: Good call on `touch: false`, I'll go ahead and add that. As for the manual update, this wouldn't erase any data that was meaningful (ideally), since only failed JSON parses would fall back to updating the model itself. If we notice people complaining of controls not being saved, we'll have to retackle #71.

PR #77, opened 2021-10-03
Properly show errors when derivatives cannot be created

CactusPuppy:
Previously when a user tried to create an invalid derivative, the form would not provide feedback on why it failed to submit. This PR alters the `Post#update` method to properly show the error(s) whenever an attempt to create invalid derivatives is made.
image


Mitcheljager (approved):
👍

PR #96, opened 2021-11-04
password reset fix

### LINKED ISSUE #94: Password reset fails for users with a `nice_url`
This is due to model validation done in https://github.com/EloHellEsports/workshop.codes/blob/629634e116096d42c2fbb7c1d66b45cba2b2e83b/app/models/user.rb#L15-L19 where users are prohibited from having an identical username to an existing person with a `nice_url` set.

However, since the user is trying to reset their own password, implying they are currently logged out, `Current.user` is `nil`, meaning they are unable to reset their password.

CactusPuppy:
Fixes #94 

TODO:
- [x] Add test against expired or invalid tokens
- [x] Add test against nonexistent user by email

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -38,7 +38,6 @@ jobs:
           sudo apt update
           sudo apt-get install libsqlite3-dev
           sudo rm config/credentials.yml.enc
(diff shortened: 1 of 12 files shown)


Mitcheljager (commented):
Very nice 👍

PR #98, opened 2021-11-23
flash rework

CactusPuppy:
- feat: Update flash display
- fix: Prevent notifications from showing on load
- feat: Tweak some styling of the alerts
- feat: Remove outdated admin dashboard flash section

### DIFF
--- a/app/assets/stylesheets/elements/_alert.scss
+++ b/app/assets/stylesheets/elements/_alert.scss
@@ -39,7 +39,7 @@
   }
 
   p {
-    max-width: 65ch;
+    max-width: 50em;
   }
 
   &--error {
--- a/app/assets/stylesheets/elements/_alert.scss
+++ b/app/assets/stylesheets/elements/_alert.scss
@@ -1,6 +1,6 @@
 @keyframes move-in-alert {
   from {
-    transform: translateY(200%);
+    transform: translateY(100vh);
   }
 
   to {
@@ -31,8 +31,7 @@
   color: $pure-white;
   font-size: 14px;
   text-decoration: none;
-  z-index: 2000;
-  transform: translateY(200%);
+  transform: translateY(100vh);
   animation: move-in-alert 200ms 500ms forwards;
   max-width: min(50em, 75%);
 
--- a/app/assets/stylesheets/elements/_alert.scss
+++ b/app/assets/stylesheets/elements/_alert.scss
@@ -8,11 +8,23 @@
   }
 }
 
-.alert {
+.alerts {
   position: fixed;
   right: $margin / 3;
   bottom: $margin / 3;
+  display: flex;
+  flex-direction: column;
+  align-items: end;
+  gap: $margin / 3;
+  width: min(35em, 75%);
+  z-index: 2000;
+}
+
+.alerts__alert {
   padding: .5em 1em;
+  display: flex;
+  align-items: center;
+  gap: 1em;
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/elements/_alert.scss
```diff
@@ -1,28 +1,46 @@
...
+  gap: $margin / 3;
+  width: min(35em, 75%);
+  z-index: 2000;
+}
+
+.alerts__alert {
   padding: .5em 1em;
+  display: flex;
+  align-items: center;
+  gap: 1em;
```
Mitcheljager, line 27: Didn't know they renamed `grid-gap` to `gap`, neat!

app/assets/stylesheets/elements/_alert.scss
```diff
@@ -1,28 +1,46 @@
...
+  transform: translateY(100vh);
   animation: move-in-alert 200ms 500ms forwards;
-  max-width: min(50em, 75%);
+
+  button {
+    background-color: transparent;
+  }
+
+  p {
+    max-width: 65ch;
```
Mitcheljager, line 42: Might be better to not use `ch` since we don't use it anywhere else

app/views/application/_flash.html.erb
```diff
@@ -1,13 +1,9 @@
...
-  </div>
-<% elsif flash[:notice] %>
-  <div class="alert">
-    <%= flash[:notice] %>
-  </div>
-<% end %>
+<div class="alerts">
+  <% flash.each do |type, message| %>
+    <% next if type == "alert" %>
+    <div class="alerts__alert<%= " alerts__alert--#{type}" if ["warning", "error"].include? type.to_s %>">
```
Mitcheljager, line 4: Can I propose to add the space in front of the `<%=` instead of in the string just for the sake of readability?

app/assets/stylesheets/utilities/_display.scss
```diff
@@ -36,3 +36,8 @@
 .faded {
   opacity: .5;
 }
+
+.fade-out {
+  opacity: 0;
+  transition: opacity 0.5s ease-out;
```
Mitcheljager, line 42: Small nitpick; Everywhere `ms` is used instead of seconds.

Mitcheljager (changes requested):
Left some small nitpicks, nothing major. Overall great job, and great addition!

app/assets/stylesheets/elements/_alert.scss
CactusPuppy: Ah, then I think `em` will probably suffice here.

app/assets/stylesheets/elements/_alert.scss
CactusPuppy: Yep, `grid-gap` is technically deprecated, since the `gap` property applies to [flex, grid, and any multi-column layout](https://css-tricks.com/almanac/properties/g/gap/).

app/views/application/_flash.html.erb
CactusPuppy: Sure, the only reason I did it this way is because the class name ends up being `"alerts__alert "` by default.

app/assets/stylesheets/utilities/_display.scss
CactusPuppy: Will fix!

Mitcheljager (approved):

PR #102, opened 2021-12-10
feat: Update success/error application views

CactusPuppy:
This makes them utilize the existing flash system, standardizing the look of the website more.

### DIFF
--- a/app/views/application/_flash.html.erb
+++ b/app/views/application/_flash.html.erb
@@ -1,4 +1,4 @@
-<div class="alerts">
+<div id="alerts" class="alerts">
   <% flash.each do |type, message| %>
     <% next if type == "alert" %>
     <div class="alerts__alert <%= "alerts__alert--#{type}" if ["warning", "error"].include? type.to_s %>">
--- a/app/views/application/_flash.html.erb
+++ b/app/views/application/_flash.html.erb
@@ -1,4 +1,4 @@
-<div id="alerts" class="alerts">
+<div class="alerts">
   <% flash.each do |type, message| %>
     <% next if type == "alert" %>
     <div class="alerts__alert <%= "alerts__alert--#{type}" if ["warning", "error"].include? type.to_s %>">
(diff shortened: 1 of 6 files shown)


app/views/application/_flash.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="alerts">
+<div id="alerts" class="alerts">
```
Mitcheljager, line 1: I'm not a big fan of ids to be honest, even though it would work perfectly fine here. To keep it consistent I'd rather use a data-role attribute. So in this case that might be `data-role="alerts"`. That way data attributes == javascript, classes == css, ids == link anchors.

app/views/application/error.js.erb
```diff
@@ -1,12 +1,29 @@
...
+  closeButton.classList.add("button", "p-0", "pl-1/16", "pr-1/16", "text-white")
+  closeButton.dataset.role = "dismiss-parent"
+  closeButton.innerText = "✕"
+
+  function dismissParent(event) {
+    const parent = this.parentNode
+    parent.classList.add("fade-out")
+    setTimeout((parent) => {
+      parent.remove()
+    }, 500, parent)
```
Mitcheljager, line 16: What does `parent` do here?

Mitcheljager (changes requested):
Over all nice! Left one comment regarding the selector.

It might be cool to transform all of this to a Svelte component at some point, and instead call an event that just adds something to an array. But that's for another day!

app/views/application/error.js.erb
CactusPuppy: The timeout function takes in one argument, a node to remove when run, and the `setTimeout` function can pass the third argument onwards in as an argument to the provided function. However, I forgot that JavaScript will include the parent function scope in the closure for the arrow function, so this is entirely unnecessary.

Mitcheljager: Ah cool, didn't know it could take more arguments!

PR #105, opened 2021-12-19
feat: Add max length validations to most form fields

### LINKED ISSUE #101: The FAQ block in the Markdown editor can cause it to repeat it's content many times over
In one case we saw a user use multiple FAQ blocks in the description of their post. All of the questions and answers in these blocks were duplicated many times over, over 200 times each.

It may or may not be related to the "Remove" button of this block not being functional, the event does not seem to be bound correctly.

The block is disabled via 9035deb4c57ae1c69ba289a7fd6c5a53eef6b0f7 for now.

Mitcheljager:
This PR adds a max length to lots of inputs that were previously unlimited. All the limits are super generous and way outside of the expected range. The largest description we currently have is for the OW2 thing, which is around 60.000 characters, most of which is HTML and CSS. I've set the limit for 100.000 characters for the post description, which is about the equivalent to 30 A4 pages full of text. Way more than anyone needs, but it would stop someone from just enter random spam in order to negatively affect the website. For example right now someone could enter a few million characters, which would blow our memory budget.
The limit for revision descriptions is a little lower at 20.000, because multiple revisions could show at a time.

I've similarly added a limit to the amount of images a post can have, which is set to 30. Once again far more than anyone would realistically need. The images they can add to the body of a post is still unlimited.

I've also added a cap to the amount of characters a block can have, which is a bit counter to what's actually happening since it's JSON. This is more so to prevent weird issues we've previously had with the FAQ block. While that doesn't necessarily solve what went wrong with that block, it does prevent it from going crazy again. Which closes #101, although it's worth investigating further.

The character limit is shown in the text editor like so;
image

If the limit is exceeded it is shown in red;
image


app/models/post.rb
```diff
@@ -119,11 +122,13 @@ class Post < ApplicationRecord
   validates :title, presence: true, length: { minimum: 5, maximum: 75 }
   validates :code, presence: true, uniqueness: { case_sensitive: false }, length: { minimum: 5, maximum: 6 }, format: { with: /\A[A-Za-z0-9]+\z/, message: "is invalid. Only letters and numbers are allowed." }
   validates :nice_url, uniqueness: true, allow_blank: true, length: { minimum: 7, maximum: 20 }, format: { with: /\A[a-z0-9-]+\z/, message: "is invalid. Only lowercase letter, numbers, and dashes are allowed." }
+  validates :description, length: { maximum: 100000 }
```
CactusPuppy, line 125: It would be nice if we could DRY these limit values out so that if we need to change them in the future, it won't be necessary to comb the entire codebase. See [Ruby on Rails: Where to define global constants?](https://stackoverflow.com/questions/4110866/ruby-on-rails-where-to-define-global-constants) for some suggestions on how to hoist this value up into a One True value.

app/models/post.rb
```diff
@@ -119,11 +122,13 @@ class Post < ApplicationRecord
...
   validates :categories, presence: true, array_length: { maximum: 3 }, array_name_part_of: { array: categories }
   validates :tags, length: { maximum: 100 }
   validates :heroes, presence: true, array_name_part_of: { array: heroes }
   validates :maps, presence: true, array_name_part_of: { array: maps }
   validates :version, length: { maximum: 20 }
+  validates :image_order, array_length: { maximum: 30 }
```
CactusPuppy, line 131: Note that for values which are only used in one place, it's definitely fine to just leave the value as-is.

CactusPuppy (changes requested):
Overall, looks great, and by my investigation, it seems we've covered most of the user-controlled inputs. Would like to DRY out some of the limit values though.

CactusPuppy (approved):
LGTM

PR #107, opened 2022-01-12
fix(webpack): "loose" config warnings

CactusPuppy:
Prevents this set of messages from appearing during the `assets:precompile` phase

```
       Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-property-in-object since the "loose" mode option was set to "true" for @babel/plugin-proposal-private-methods.

       The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding

       	["@babel/plugin-proposal-private-property-in-object", { "loose": true }]

       to the "plugins" section of your Babel config.
```

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -16,6 +16,10 @@ module.exports = function(api) {
   }
 
   return {
+    assumptions: {
+      privateFieldsAsProperties: true,
+      setPublicClassFields: true
+    },
     presets: [
       isTestEnv && [
         '@babel/preset-env',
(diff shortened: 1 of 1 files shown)


Mitcheljager (approved):
Eh, fair enough. Sometimes it's easier to leave these files alone, but why not.

PR #118, opened 2022-02-24
fix: Several profile fixes

Mitcheljager:
This PR fixes several things related to the profile pages:

- Fix incorrect placement of the favorite button in profile highlight
- Decrease border radius in profile highlight
- Fix 404 when post no longer exists in profile highlight
- Fix infinite scroll loading even if the posts are not visible in the tab

## Highlight before/after
image
image


CactusPuppy (approved):
LGTM!

PR #119, opened 2022-02-25
fix: Rename banner class to standout to fix adblocker blocking the element

Mitcheljager:
The element would get blocked by some adblockers because `.banner` is identified as an ad. This fixes that by renaming the class to `.standout` (naming is hard ok)


CactusPuppy:
My day be so fine, then suddenly `.banner` is not allowed 😔

CactusPuppy (approved):
LGTM!

PR #130, opened 2022-03-29
feat: Add April Fools' Joke

CactusPuppy:
Adding a small, fun change for April Fools' this year.﻿

### DIFF
--- a/app/javascript/src/april-fools.js
+++ b/app/javascript/src/april-fools.js
@@ -1,0 +1,21 @@
+export function destroy() {
+  const elements = document.querySelectorAll("[data-role~='april-fools']")
+
+  elements.forEach((element) => element.parentNode.remove())
+}
+
+export function bind() {
+  // Pick a random number between 1 and 4, and if it's 1, or the user sets the `april-fools` parameter, show the April Fools joke.
+  const randomNumber = Math.floor(Math.random() * 4) + 1
+  const url = new URL(window.location.href)
+  if (randomNumber !== 1 && !url.searchParams.get("april-fools")) return
+
+  const alertsArea = document.querySelector("[data-role='alerts']")
+
+  const foolsAlert = document.createElement("div")
+  foolsAlert.dataset.role = "april-fools"
+  foolsAlert.classList.add("alerts__alert", "alerts__alert--warning")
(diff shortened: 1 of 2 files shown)


app/javascript/src/april-fools.js
```diff
@@ -0,0 +1,27 @@
+import FetchRails from "./fetch-rails"
+
+export function destroy() {
+  const elements = document.querySelectorAll("[data-role~='april-fools']")
+
+  elements.forEach((element) => element.parentNode.remove())
+}
+
+export function bind() {
```
Mitcheljager, line 9: You're not really binding anything, so techncailly it could just be `document.addEventListener("turbolinks:load", function() {`, but I'm not too concerned about it.

app/javascript/src/april-fools.js
```diff
@@ -0,0 +1,27 @@
...
+    const randomNumber = Math.floor(Math.random() * chance) + 1
+    const url = new URL(window.location.href)
+    if (randomNumber !== 1 && !url.searchParams.get("april-fools")) return
+
+    const alertsArea = document.querySelector("[data-role='alerts']")
+
+    const foolsAlert = document.createElement("div")
+    foolsAlert.dataset.role = "april-fools"
+    foolsAlert.classList.add("alerts__alert", "alerts__alert--warning")
+    foolsAlert.innerHTML = "<p class=\"mt-1/16 mb-1/16\">The server closed due to excessive Workshop script load. <a class=\"text-pure-white\" title=\"Click for explanation\" href=\"/explanation\" data-prefetch=\"false\">(?)</a></p><button name=\"button\" type=\"submit\" class=\"button p-0 pl-1/16 pr-1/16 text-white\" data-role=\"dismiss-parent\">✕</button>"
```
Mitcheljager, line 23: You could use \` \` template literals to avoid the escaping of `"`

Mitcheljager (approved):
Left 2 comments, but to be honest I don't really care, it's just for fun anyway. If you want to fix them, do, otherwise feel free to merge!

app/javascript/src/april-fools.js
CactusPuppy: I did this because while no event gets bound, it is still necessary to ensure that this happens before `dismissParent` fires so that the close button is bound. I will change the function name to avoid confusion though.

app/javascript/src/april-fools.js
CactusPuppy: ESLint "fixed" this "styling problem" automatically for me. I'll revert the "fix" and tell ESLint to be quiet.

app/javascript/src/april-fools.js
Mitcheljager: Ah fair enough!

PR #135, opened 2022-04-02
Lighthouse improvements

Mitcheljager:
This PR improves a few warnings from Lighthouse, I want to get back to 100%;

- Added passive to touch events from Siema slider. This is done through a patch, but unfortunately Siema only ships the package minified, so the entire file is patched. Kinda sucks, but it's better than not having passive events.
- The NoUiSlider used for player counts adds a fair amount of dom depth and elements. The one in the filter will now only be rendered when you open the filter. This is done through an observer.
- No longer inline the Favorite icon svg. Technically this saves about ~40kb from the root document, but in reality it's compressed anyway so it doesn't save anything. It saves some dom elements, it can be cached, and maybe it saves some render time for the server, but that's probably negligible.
- Remove unused CSS and move logged in user specific css to their respective file. This saves about 60kb (~280kb to ~220kb) before minifying and compression.


CactusPuppy (approved):
LGTM!

PR #137, opened 2022-04-11
fix: Information leaks with derivation system

### LINKED ISSUE #133: [Bug] Unlisted post creation sends notification to source post author
A notification is incorrectly sent to the author of a which an unlisted post derives from when the unlisted post is created.

### LINKED ISSUE #134: [Bug] Unlisted posts appear in the derivations of a source post
Unlisted posts appear in the derivations section of a post the unlisted post sources from

CactusPuppy:
Fixes #133 and fixes #134


features/derivatives.feature
```diff
@@ -0,0 +1,66 @@
+@javascript
+Feature: Post derivatives
+
+  As a post creator on Workshop.codes
+  So that I can credit other people's work in my post
+  I want to be able to add sources to my post
+
+  Background: Two users and two posts each
+    Given a user named "OBSWinston"
```
Mitcheljager, line 9: Code aside, the OWL memes are lovely :D

features/step_definitions/derivative_stepdefs.rb
```diff
@@ -0,0 +1,87 @@
...
+  fail "Need to be on a post edit page" unless page.current_path.ends_with?("/edit")
+  post = Post.find_by_title(title)
+  expect(post).to be_present
+
+  within "[data-svelte-component='DerivativesForm']" do
+    close = find("span", text: post.code).find("div")
+    close.click
+  end
+
+  click_on 'Save'
```
Mitcheljager, line 37: Should be double quotes

app/models/post.rb
```diff
@@ -183,6 +183,10 @@ def self.public?
     self.where(private: false, unlisted: false, draft: false)
   end
 
+  def public?
```
Mitcheljager, line 186: Won't this interfere with the definition above? Or are they separate because one is a query and the other is for individual posts?

https://github.com/EloHellEsports/workshop.codes/blob/552091aca2207c4645724ef8e84b3b3bcb02ff42/app/models/post.rb#L182-L184

app/controllers/application_controller.rb
```diff
@@ -39,7 +39,7 @@ def reject_if_banned
 
   def current_user
     if session[:user_id]
-      @current_user ||= User.find(session[:user_id])
+      @current_user ||= User.find_by(id: session[:user_id])
```
Mitcheljager, line 42: Did you change this so it 404's? If so, is that better than 500?

Mitcheljager (approved):
Asked some question and left some minor comments, but overall great job!

features/step_definitions/derivative_stepdefs.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  post = Post.find_by_title(title)
+  expect(post).to be_present
+
+  check id: "show_derivative", allow_label_click: true
+
+  # We bypass the normal source selection by using the hidden field
+  page.evaluate_script "document.getElementById('post_derivatives').value = '#{ post.code }'"
+
+  click_on 'Save'
```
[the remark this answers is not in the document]
CactusPuppy: Rubocop when /s

app/models/post.rb
CactusPuppy: They are separate indeed. `self.public` refers to the class, and is a method to filter a query. `public?` is a method on an instance of the class, and refers only to a particular instance.

app/controllers/application_controller.rb
CactusPuppy: I changed this because it is apparently possible for a Remember token to remain valid even after the deletion of a user, which was causing some tests to malfunction by attempting to log in as a non-existent user. After typing the previous sentence out, I think we should probably do something about that.

Mitcheljager: Hmm remember tokens are destroyed when the user is destroyed, I wonder what's going on there

CactusPuppy: If that's the case, maybe there's something fishy happening with the session token itself being retained? We can chat more on Discord for faster communication.

CactusPuppy: Resolved on Discord. Issue being tracked in #140

PR #142, opened 2022-04-15
feat: Lazy load similar codes when scrolled to

Mitcheljager:
This PR makes it so the Similar Codes on post/show only load if they are actually scrolled to. This is to prevent unnecessary calls to ElasticSearch for something that may never get seen.

You can now pass `data-lazy="true"` to an element with `data-action="get-partial"` to lazy load it on scroll. This same function can be used to either load immediately on page load or load on click. It's currently only used for these similar codes, so the other functions are a little redundant, but might as well keep them for some other time.

### DIFF
--- a/app/javascript/src/get-partial.js
+++ b/app/javascript/src/get-partial.js
@@ -4,19 +4,33 @@ export function bind() {
   const elements = document.querySelectorAll("[data-action~='get-partial']")
 
   elements.forEach(element => {
-    element.removeAndAddEventListener("click", getPartial)
-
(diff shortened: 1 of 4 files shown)


CactusPuppy:
Might be nicer to use the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) over a traditional scroll listener. It has [good support across modern browsers](https://caniuse.com/?search=intersection%20observer) and reduces load on clients.

Mitcheljager:
@CactusPuppy I like that! I will give it a go.

PR #143, opened 2022-04-15
fix: Handle errors when favoriting

Mitcheljager:
Sometimes when favoriting or removing a favorite the post can not be found. I'm not sure why this happens, but it does. This now brings up an alert showing that something went wrong.

image

I've also slightly tweaked the styling of the close button of the alert, making it a little smaller.

image


CactusPuppy (approved):
LGTM

PR #144, opened 2022-04-23
feat: Overwatch 2 beta indicator

Mitcheljager:
This PR adds an option to indicate a post as being OW2 compatible. We're not yet 100% sure what the Workshop in the beta will look like and if codes carry over between live or different beta phases, but this PR serves as a base that we can change later when we know more.

The setting to indicate a post as being for the PTR has been replaced with a switch to toggle OW2. The PTR flag is still available, it's just hidden, as it might return in the future.

Posts themselves have a flag similar the to PTR flag as well as a banner on the post show itself.

The root has a standout with a direct link to a page showing only OW2 codes. This is just the filter page so any other filters can also be applied (including search terms).

The filter page has another banner showing some extra info on what to expect from codes between OW1 and OW2 betas.

There are two routes, one is the basic filter which is `/overwatch-2/true/`, the other is an alias which is just `/overwatch-2`, which is a bit nicer for SEO.

I've added some really basic tests using the default rails test.

This also adds the first new maps and hero that will be available in the beta. These both are clearly marked as being OW2 specific in the UI.

## Switch in the settings
| Default | Toggled |
| --- | --- |
| image | image |

## Indicator on post
image

## Banner on homepage
| Desktop | Mobile |
| --- | --- |
| image | image |

## Filter page indicator (only visible of OW2 is selected)
image

## Toggle in the filter
image

## Banner on post/show
image

## Tooltip in banner
The text doesn't flow very nicely, any suggestions for different wording?
image

## Dropdowns show OW2 specific heroes and maps with an OW2 marker
image

## Same for in the settings of a post
image


app/views/filter/index.html.erb
```diff
@@ -24,4 +25,20 @@
...
+  <div class="well well--dark b-overwatch-2 mb-1/2">
+    <div>
+      <h3 class="mt-0 mb-0">
+        <strong><%= t("banner.overwatch_2.title_html") %></strong>
+      </h3>
+
+      <div class="mt-1/4 text-small">
+        These Workshop Codes are verified to be compatible with the Overwatch 2 Beta.
+        There is no guarantee that codes will function between different beta phases.
+        A code that works in the Overwatch 2 Beta will not work in Overwatch 1.
```
CactusPuppy, line 38: This wording is slightly confusing. I'd suggest something like the following:
```suggestion
        These Workshop codes were tagged as being importable within Overwatch 2.
        It is possible that codes may have been invalidated, since Overwatch 2 has not been fully released yet.
        An import code that was created in Overwatch 2 will not work in Overwatch 1.
```

app/views/posts/_overwatch_2_indicator.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="warning warning--orange bg-overwatch-2">
+  <div>
+    <strong>Overwatch 2</strong> Beta Compatible
+
+    <em class="tooltip tooltip--right">
+      (?)
+      <div class="tooltip__content">
+        This code is verified to be compatible with the Overwatch 2 Beta.
+        There is no guarantee that codes will function between different beta phases.
+        A code that works in the Overwatch 2 Beta will not work in Overwatch 1.
```
CactusPuppy, line 10: Similar to the rewording of the Overwatch 2 index page, I'd suggest something like:
```suggestion
        This code is tagged as being importable in Overwatch 2.
        It is possible the code was only valid for a previous beta, or that functionality has broken due to updates.
        A code that works in the Overwatch 2 Beta will not work in Overwatch 1.
```

config/locales/en.yml
```diff
@@ -73,6 +73,9 @@ en:
       tagline: Turns out websites are expensive and ads suck
       content_html: Support Workshop.codes for just <span class="text-white">$3</span>
       button: Support Workshop.codes
+    overwatch_2:
+      title_html: <span class="text-overwatch-2">Overwatch 2</span> Beta compatible&nbsp;codes
```
CactusPuppy, line 77:
```suggestion
      title_html: <span class="text-overwatch-2">Overwatch 2</span> compatible&nbsp;codes
```

config/locales/en.yml
```diff
@@ -73,6 +73,9 @@ en:
...
+      button: Check it out
```
CactusPuppy, line 78:
```suggestion
      button: View
```

CactusPuppy (changes requested):
One specific note, since I can't comment on binary files: would we prefer to have an SVG, or a higher resolution Overwatch 2 logo? I can make any of those happen if need be.

Other than that, some wording suggestions were included. All the technical stuff looks good (including tests!)

config/locales/en.yml
Mitcheljager: I kept Beta in here for SEO. I'm assuming people might search google for "overwatch 2 beta workshop" or something along those lines.

CactusPuppy: Ah ok, then that's fine

app/views/filter/index.html.erb
```diff
@@ -24,4 +25,20 @@
...
+  <div class="well well--dark b-overwatch-2 mb-1/2">
+    <div>
+      <h3 class="mt-0 mb-0">
+        <strong><%= t("banner.overwatch_2.title_html") %></strong>
+      </h3>
+
+      <div class="mt-1/4 text-small">
+        These Workshop codes were tagged as being importable within Overwatch 2.
+        It is possible that codes may have been invalidated, since Overwatch 2 has not been fully released yet.
+        An import code that was created in Overwatch 2 will not work in Overwatch 1.
```
CactusPuppy, line 38:
```suggestion
        These Workshop codes were tagged as being importable within the Overwatch 2 Beta.
        It is possible that codes may have been invalidated, since the Overwatch 2 Beta is under active development.
        An import code that was created in Overwatch 2 Beta will not work in Overwatch 1.
```

app/views/posts/_overwatch_2_indicator.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="warning warning--orange bg-overwatch-2">
+  <div>
+    <strong>Overwatch 2</strong> Beta Compatible
+
+    <em class="tooltip tooltip--right">
+      (?)
+      <div class="tooltip__content">
+        This code is tagged as being importable in Overwatch 2.
+        It is possible the code was only valid for a previous beta, or that functionality has broken due to updates.
+        A code that works in the Overwatch 2 Beta will not work in Overwatch 1.
```
CactusPuppy, line 10:
```suggestion
        This code is tagged as being importable in the Overwatch 2 Beta.
        It is possible the code was only valid for a previous beta, or that functionality has broken due to Overwatch 2 Beta updates.
        A code that works in the Overwatch 2 Beta will not work in Overwatch 1.
```

CactusPuppy (commented):
I've suggested a few more tweaks to the copy for SEO purposes, but honestly I could take them or leave them.

CactusPuppy (approved):
LGTM!

PR #146, opened 2022-04-27
feat: Add graceful UX for search page errors

CactusPuppy:
Previously, if Elasticsearch returned an error during search, it would fail silently. This PR explains the issue to the user and offers a retry option.

This button appears whenever Load More or Infinite Scroll fails to load more posts via Ajax:
image

Pagination users will be given an error page:
image

### DIFF
--- a/app/controllers/filter_controller.rb
+++ b/app/controllers/filter_controller.rb
@@ -24,7 +24,7 @@ class FilterController < ApplicationController
       @posts = Kaminari.paginate_array(@posts).page(params[:page])
     rescue Elasticsearch::Transport::Transport::ServerError => e
       Bugsnag.notify(e) if Rails.env.production?
-      @posts = []
+      @posts = Kaminari.paginate_array([]).page(params[:page])
       @message = "Something went wrong. Please try again later."
       flash[:error] = @message
       respond_to do |format|
--- a/app/controllers/filter_controller.rb
+++ b/app/controllers/filter_controller.rb
@@ -1,6 +1,5 @@
 class FilterController < ApplicationController
   def index
-    # binding.pry
     begin
       @posts = params[:search] ? Post.includes(:user).where(id: Post.search(params[:search], bypass_cache: true)).select_overview_columns.public? : Post.select_overview_columns.public?
 
--- a/app/controllers/filter_controller.rb
+++ b/app/controllers/filter_controller.rb
@@ -28,7 +28,7 @@ class FilterController < ApplicationController
       @message = "Something went wrong. Please try again later."
       flash[:error] = @message
       respond_to do |format|
-        format.html { render :index, status: 500 }
+        format.html { render "errors/internal_error", status: 500 }
         format.js { render "posts/infinite_scroll_posts", status: 500 }
         format.json { render json: { message: @message }, status: 500 }
       end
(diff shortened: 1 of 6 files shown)


Mitcheljager:
Hmm I'm not sure how I feel about showing the 500 error directly in the layout. It's better of course since you can search again, but it doesn't look great. Maybe it needs some extra styling, or maybe it needs to be a separate design. This is more work, but just brainstorming; It could be a page that says "Something went wrong with searching, try again", with a big-ass search box right under it. As it is the "Return home" button as a fallback seems irrelevant as the logo button and navigation are there anyway.

CactusPuppy:
> Hmm I'm not sure how I feel about showing the 500 error directly in the layout. It's better of course since you can search again, but it doesn't look great. Maybe it needs some extra styling, or maybe it needs to be a separate design. This is more work, but just brainstorming; It could be a page that says "Something went wrong with searching, try again", with a big-ass search box right under it. As it is the "Return home" button as a fallback seems irrelevant as the logo button and navigation are there anyway.

I hear you. It probably wouldn't be too much work to just tweak the search results partial now that I think about it.

app/controllers/filter_controller.rb
```diff
@@ -1,25 +1,38 @@
...
+      @posts = @posts.select { |post| to_slug(post.categories).include?(to_slug(params[:category])) } if params[:category]
+      @posts = @posts.select { |post| to_slug(post.maps).include?(to_slug(params[:map])) } if params[:map]
+      @posts = @posts.select { |post| to_slug(post.heroes).include?(to_slug(params[:hero])) } if params[:hero]
+      @posts = @posts.select { |post| helpers.has_player_range?(post) && to_range(params[:players]).overlaps?((post.min_players)..(post.max_players)) } if params[:players]
+      @posts = @posts.select { |post| post.code.upcase.start_with?(params[:code].upcase) } if params[:code]
 
-    @posts = Kaminari.paginate_array(@posts).page(params[:page])
+      @posts = Kaminari.paginate_array(@posts).page(params[:page])
+    rescue Elasticsearch::Transport::Transport::ServerError => e
+      Bugsnag.notify(e) if Rails.env.production?
```
Mitcheljager, line 25: 👍

app/javascript/src/infinite-scroll.js
```diff
@@ -64,6 +64,20 @@ function getInfiniteScrollContent(element) {
         element.innerHTML = "Load more"
         element.setAttribute("data-url", nextUrl)
       }
+    },
+    error: (error) => {
+      progressBar.setValue(1)
+      progressBar.hide()
+
+      const spinner = document.querySelector(".items").querySelector(".spinner")
```
Mitcheljager, line 72: Let's add some optional chaining here, just in case.

`document.querySelector(".items")?.querySelector(".spinner")`

app/javascript/src/infinite-scroll.js
```diff
@@ -64,6 +64,20 @@ function getInfiniteScrollContent(element) {
...
+      if (spinner) spinner.remove()
+      if (element.classList.contains("button")) {
```
Mitcheljager, line 74: Not the nicest way of checking this, if we change the style of the button (and with that the class name) this will fail, but meh, that's a problem for another day

Mitcheljager (changes requested):

app/javascript/src/infinite-scroll.js
CactusPuppy: Shouldn't be too much work to add an extra `data-*` attribute to distinguish the two elements. I'll take care of that now to avoid technical debt.

PR #149, opened 2022-05-09
User session cookie improvements

### LINKED ISSUE #140: [Bug] Session token is not able to be invalidated
In certain circumstances, such as when a user is destroyed on another device, a session cookie may still prompt the `ApplicationController` to attempt to log in as an invalid user (or worse, as the wrong user).

https://github.com/EloHellEsports/workshop.codes/blob/7d0cd03da2d139137643cdf40e49e2d18767b695/app/controllers/application_controller.rb#L41-L43

This risk is currently tolerable for this project, since the session cookie is encrypted and infeasible to forge with a set value without compromising the server's production key. However, it still is important to fix.

CactusPuppy:
This PR adds a UUID to the User model which is used to ensure that a session cannot incorrectly log a user into an account they do not own, which fixes #140. For user convenience, legacy sessions will continue to function.

This PR also includes a countermeasure to prevent [session fixation attacks](https://edgeguides.rubyonrails.org/security.html#session-fixation).

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -66,9 +66,12 @@ class User < ApplicationRecord
                             dimension: { max: 3500..3500 }
   validates :uuid, presence: true, uniqueness: true, length: { is: 36 }, format: { with: /\A[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\z/i }
 
-  before_validation do
-    self.uuid = SecureRandom.uuid unless self.uuid.present?
+  after_find do
+    # binding.pry
+    self.update_column(:uuid, SecureRandom.uuid) unless self.uuid.present?
   end
+
+  before_validation :generate_and_set_uuid, on: :create
 
   def self.find_or_create_from_auth_hash(auth_hash)
     uid = auth_hash["uid"]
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -289,6 +289,7 @@
     t.text "custom_css"
     t.integer "pagination_type", default: 0
     t.integer "linked_id"
+    t.uuid "uuid", default: -> { "gen_random_uuid()" }
```
Mitcheljager, line 292: I thought SQLite doesn't support UUIDs, or am I wrong?

CactusPuppy: That is correct. I have PostgreSQL set up locally, so I didn't catch that snag.

Mitcheljager: I've wanted to replace all IDs with UUIDs, but SQLite is so nice in dev, don't want to give it up

app/models/user.rb
```diff
@@ -64,6 +64,14 @@ class User < ApplicationRecord
   validates :banner_image, content_type: ["image/jpeg", "image/jpg", "image/png"],
                             size: { less_than: 2.megabytes },
                             dimension: { max: 3500..3500 }
+  validates :uuid, presence: true, uniqueness: true, length: { is: 36 }, format: { with: /\A[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\z/i }
+
+  after_find do
```
Mitcheljager, line 69: Never seem this before, kinda weird and kinda cool!

Mitcheljager (approved):
Nicely done :)

PR #150, opened 2022-05-11
fix: Ensure pagination works within profiles

### LINKED ISSUE #83: [BUG] Pagination setting conflicts with profile redesign
**Description:** If User A has blocks directly referencing posts, and User B tries to visit the next page of codes created, an error will occur.

**Steps to Reproduce:**
1. Find or set up a user profile which has blocks referencing posts directly as well as enough codes to spill into a new page with Pagination type set to Pagination.
2. Visit Accessibility settings page.
3. Set Pagination type to Pagination.
3. Visit user profile from Step 1.
4. Open Codes section and scroll to the bottom.
5. Attempt to visit the next page.

**Expected behavior**: Browser remains in the Codes tab and displays the next page of codes from this user.

**Actual behavior**: An error is encountered.

CactusPuppy:
This PR fixes #83

### DIFF
--- a/app/controllers/posts_controller.rb
+++ b/app/controllers/posts_controller.rb
@@ -119,7 +119,8 @@ class PostsController < ApplicationController
         end
         return
       end
-      @revision = Revision.new(post_id: @post.id, code: @post.code, version: @post.version, snippet: @post.snippet).save
+      @revision = Revision.new(post_id: @post.id, code: @post.code, version: @post.version, snippet: @post.snippet)
+      @revision.save
 
       create_activity(:create_post, post_activity_params)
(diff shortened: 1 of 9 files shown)


Mitcheljager (approved):
Looks good!  One question; why is `.except(:limit, :offset)` necessary? I'm not sure I know what's going on

CactusPuppy:
> Looks good! One question; why is `.except(:limit, :offset)` necessary? I'm not sure I know what's going on

@Mitcheljager this is because profile blocks only store the ID of the post, then pull the post from the `@posts` variable. However, because said variable is paginated, the desired post may not be accessible from the current page.

CactusPuppy:
BLOCKING: Need RSpec tests to cover bug

PR #156, opened 2022-06-04
feat: Update post analytics with impressions count

Mitcheljager:
This PR adds the total impressions count to post analytics. It also adds cards similar to the analytics on the user overview page. I also wanted to add total copies count but we don't have a cache counter for that yet so I will leave that for a separate PR.

## User posts page
| Desktop | Mobile |
| --- | --- |
| image |  image |

## Analytics on post show
image


CactusPuppy (approved):
LGTM!

PR #157, opened 2022-06-04
feat: Scroll anchor for comment notifications

Mitcheljager:
This PR adds a scroll anchor to comment notifications so it scrolls to the correct comment on load. Since the comments are loaded async this required adjusting the `getPartial` function to scroll after loading. The added benefit of this is that the scroll is smooth rather than instant, which looks neat. I've also refactored parts of this function to catch errors and show them, just in case.

### DIFF
--- a/app/javascript/src/get-partial.js
+++ b/app/javascript/src/get-partial.js
@@ -34,7 +34,26 @@ function getPartial(event, element) {
 
   new FetchRails(url).get()
     .then(data => {
-      targetElement.dataset.loaded = "true"
       targetElement.innerHTML = data
+    })
+    .then(() => {
+      if (_this.dataset.scrollOnLoad != "true") return
+
+      const hash = window.location.hash?.substring(1)
+      if (!hash) return
+
+      const scrollElement = document.getElementById(hash)
+      if (!scrollElement) return
+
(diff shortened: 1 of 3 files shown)


app/javascript/src/get-partial.js
```diff
@@ -34,7 +34,26 @@ function getPartial(event, element) {
...
+
+      const hash = window.location.hash?.substring(1)
+      if (!hash) return
+
+      const scrollElement = document.getElementById(hash)
+      if (!scrollElement) return
+
+      const scrollOffset = scrollElement.getBoundingClientRect().top + document.documentElement.scrollTop
+
+      window.scrollTo({ top: scrollOffset - 70, behavior: "smooth" })
```
CactusPuppy, line 50: Minor nit: it would nice to have a comment or something explaining this "magic" number of 70, although I could hazard a guess that this was due to the sticky header plus visual space.

CactusPuppy (changes requested):
One minor documentation thing, then I think we're ready to roll this out!

app/javascript/src/get-partial.js
Mitcheljager: You're right, I got lazy and used a static number. I've replaced it with using the actual sticky tabs height so we don't have to think about this if we ever change those tabs or remove them altogether.

PR #161, opened 2022-06-17
feat: Show OW2 banner only if env var is set

Mitcheljager:
This PR makes it so we can set the OW2 banner on the homepage conditionally with an env variable. I decided to go with a general variable for the banner so we could potentially use it for other banners later down the line.


CactusPuppy:
LGTM!

PR #162, opened 2022-06-18
chore: Remove Google Analytics

Mitcheljager:
This PR removes Google Analytics and all of it's components. At the moment we don't use it much and we're better off with the increased speed, better user experience, and without selling out our users.


CactusPuppy (approved):
LGTM!

PR #171, opened 2022-07-28
refactor: use query parameters instead of path-based parameters

### LINKED ISSUE #165: [Refactor] Allow searches to use standard URL query parameters
Currently searches on Workshop.codes are formatted as
```
/overwatch-2/true/maps/kings-row/search/test
```
However, this is not the standard method for specifying query parameters. Typically, parameters are specified as follows:
```
/search?q=test&overwatch-2=true&maps=kings-row
```
In addition to being more implicitly understood by SEO, this change would also enable easier API integration with Workshop.codes and fewer restrictions on what values can be utilized as parameters.

CactusPuppy:
As detailed in #165, query parameters are easier to work with for applications integrating with Workshop.codes, and may be better understood by SEO algorithms.

Resolves #165

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -28,7 +28,9 @@ class SearchController < ApplicationController
       redirect_to root_path
       return
     end
+
     track_action
+
     begin
       @posts = get_filtered_posts(params)
     rescue Elasticsearch::Transport::Transport::ServerError => e
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -4,11 +4,22 @@ class SearchController < ApplicationController
   def index
     unless params[:query].empty?
       respond_to do |format|
-        format.js { redirect_to build_filter_path(:search, params[:query].gsub(".", "")) }
-        format.html { redirect_to build_filter_path(:search, params[:query].gsub(".", "")) }
+        format.js { redirect_to build_filter_path(:search, params[:query]) }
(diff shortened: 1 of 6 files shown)


app/controllers/search_controller.rb
```diff
@@ -3,19 +3,55 @@ class SearchController < ApplicationController
...
+    end
+  end
+
   def show
-    redirect_to root_path
+    if params.keys.without("controller", "action", "default").length == 0
+      redirect_to root_path
+      return
+    end
+    track_action
```
Mitcheljager, line 31: Giant nitpick but could you add some white space around `track_action`?

app/controllers/search_controller.rb
```diff
@@ -3,19 +3,55 @@ class SearchController < ApplicationController
...
-        format.html { redirect_to build_filter_path(:search, params[:query].gsub(".", "")) }
+        format.js { redirect_to build_filter_path(:search, params[:query]) }
+        format.html { redirect_to build_filter_path(:search, params[:query]) }
       end
     else
       redirect_back fallback_location: root_path
     end
   end
 
+  def old_index
```
Mitcheljager, line 15: Maybe we could come up with a better name. `redirect_to_param` or something, not entirely sure. Or add a comment what this is.

Mitcheljager (approved):
Really nice solution over all, much simpler than I thought it might have to be! I left 2 nit picks, but great job :)

PR #172, opened 2022-07-28
chore: Avoid deprecation warning for encrypted email

CactusPuppy:
Due to [this change in Lockbox](https://github.com/ankane/lockbox/commit/7477de8bfba5c7f0bf3d77c1cafd85e9411ae92b), it is necessary to rename `encrypts` to `has_encrypted`.


Mitcheljager (approved):
Good catch!

PR #173, opened 2022-07-31
feat: Improve focus states for several form elements

Mitcheljager:
This PR improves the `focus` and `focus-visible` states for several elements, mainly form elements. `focus-visible` is only visible when using keyboard navigation. This focus should always be clearly visible and the way it looks is far less important than the functionality.

focus-visible
focus-visible-2

The regular focus state for form elements was previously non existent. I've added a slight highlight and outline to inputs.
image
image

### DIFF
--- a/app/assets/stylesheets/_mixins.scss
+++ b/app/assets/stylesheets/_mixins.scss
@@ -63,3 +63,23 @@
     outline: none;
   }
 }
+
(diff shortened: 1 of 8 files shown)


CactusPuppy (commented):
Looks great for the most part! I'd say we could add a `border-radius` on the header navigation links, but other than that, this is a great addition to the site.

PR #177, opened 2022-08-02
fix: Search filters not properly applying

CactusPuppy:
Due to changes made in #171, the previous `filter.js` improperly named its query parameters, meaning the filters did not function. This PR rectifies that.

Examples:
`categories=___` -> `category=___`
`maps=___` -> `map=___`

### DIFF
--- a/app/controllers/filter_controller.rb
+++ b/app/controllers/filter_controller.rb
@@ -1,7 +1,7 @@
 class FilterController < ApplicationController
   def index
     # REFACTOR: Make this set of attributes have a single source of accessible truth
-    filtered_params = params.permit([:category, :hero, :map, :from, :to, :sort, :expired, :author, :players, :language, :search, :default, :page])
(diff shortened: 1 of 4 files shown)


app/controllers/filter_controller.rb
```diff
@@ -1,7 +1,7 @@
 class FilterController < ApplicationController
   def index
     # REFACTOR: Make this set of attributes have a single source of accessible truth
-    filtered_params = params.permit([:category, :hero, :map, :from, :to, :sort, :expired, :author, :players, :language, :search, :default, :page])
```
Mitcheljager, line 4: What was `:default` even?

Mitcheljager (approved):
Good catch! I had one question, but feel free to merge!

app/controllers/filter_controller.rb
CactusPuppy: Default locale I believe. Either way, it gets added with the redirected request again, so it is not necessary to include.

PR #178, opened 2022-08-02
refactor: Rename search query parameter

CactusPuppy:
To avoid confusion between `params[:search]` and `params[:query]`, this PR changes all uses of the searched query to be `:query`.


CactusPuppy:
Upon further review, this is probably entirely unnecessary. Closing to focus resources elsewhere.

PR #184, opened 2022-08-11
chore: Update robots.txt with Yandex and PetalBot

Mitcheljager:
This PR updates robots.txt to disallowed Yandex and PetalBot. PetalBot is a chinese bot from Huawei. We have no interest in chinese traffic as our codes won't work anyway. Yandex is Russian, and while a somewhat significant portion of traffic comes from Russia, I see no point in allowing them to index us.

I've also added a crawl-delay (I thought we had this already, but I guess not). This will prevent future bots from crawling too aggressively, as long as they respect it of course.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,5 +1,8 @@
 Sitemap: https://workshop.codes/sitemap.xml
 
+User-agent: *
+Crawl-delay: 5
+
 User-agent: SemrushBot
 Disallow: /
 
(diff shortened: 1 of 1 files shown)


CactusPuppy (approved):
Approved as-is, but for consideration: should we see if PetalBot and Yandex respect the crawl-delay before outright blocking them? It's up to you either way.

Mitcheljager:
@CactusPuppy Good point but to be honest I don't really care for either Yandex or PetalBot.

PR #186, opened 2022-08-12
fix: Prevent undefined behavior with blank comments

CactusPuppy:
This PR fixes a few minor issues with comment handling. In particular:
- The comments controller did not properly respond with an error due to a lack of an enclosing `respond_to` block
- The comments form did not contain a `required` attribute, which led to poor user feedback.

This PR also contains regression tests to capture the current behavior


Mitcheljager (approved):
Nice!

PR #187, opened 2022-08-12
test: Prevent rare test interdependency

CactusPuppy:
Due to a rare test ordering bug, it was possible, though unlikely, for testing of the `application_helper`'s i18n values to cause a test interdependency with the profiles specs. This PR addresses this concern.

Note: Issue was first noticed with `bundle exec rspec --seed 17258` in [this CI run](https://github.com/CactusPuppy/workshop.codes/runs/7799224654?check_suite_focus=true)


Mitcheljager (approved):
Ahh good catch, I wouldn't have thought of that!

PR #189, opened 2022-08-18
chore: Remove simple-view accessibility setting

Mitcheljager:
The simple view is super broken and the site is unusable with it turned on. Maintaining it is not worth the effort as hardly anyone ever used it and it's purpose isn't clear. What it is trying to do should instead be done via other ways as having to toggle it isn't a good experience.

### DIFF
--- a/app/views/posts/index.html.erb
+++ b/app/views/posts/index.html.erb
@@ -3,7 +3,7 @@
 
 <main class="top-offset">
   <% unless params[:page] %>
-    <h2 class="mt-1/2 <%= "text-black" unless current_user.present? && current_user.simple_view? %>">
+    <h2 class="mt-1/2">
       <%= inline_svg_tag "icons/icon-onfire.svg", class: "title-icon" %>
       <%= t(".on_fire.title_html") %>
     </h2>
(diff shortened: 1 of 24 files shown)


app/views/posts/index.html.erb
```diff
@@ -3,7 +3,7 @@
 
 <main class="top-offset">
   <% unless params[:page] %>
-    <h2 class="mt-1/2 <%= "text-black" unless current_user.present? && current_user.simple_view? %>">
+    <h2 class="mt-1/2">
```
CactusPuppy, line 6: Removing the `text-black` CSS class causes the On Fire header to fail the WCAG contrast test.
image

```suggestion
    <h2 class="mt-1/2 text-black">
```
image

CactusPuppy (changes requested):
Overall, great change. Requesting one change to the homepage, and then it's good to merge.

app/views/posts/index.html.erb
Mitcheljager: Ah I didn't intend to remove that anyway, I read the if statement backwards!

PR #190, opened 2022-08-18
feat: User based and post to post recommendations

Mitcheljager:
This PR uses the awesome https://github.com/ankane/disco gem to add recommendations.

There are two types of recommendations;
 1. User recommendations. These are general recommendations for posts based on your favorites and the favorites of other users. These show on the homepage for users who have favorites and recommendations based on those favorites.
 2. Post to post recommendations. These are recommendations for a single post based on users who have favorited this and other posts. These show on post show pages only if the post has any favorites and has recommendations. **It might turn out that this creates too many db records, I'm really not sure. If this is the case I might end up removing these recommendations.**
 
Both of these recommendations are only based on what favorites they have in common, they aren't based on tags, titles, or whatever else. Because of this they stand out from "Similar codes" on post pages, which are purely based on tags.
 
The recommendations are generated via a rake task. This rake task will be configured to run every x amount of time. I'm currently thinking every hour, similar to setting hotness for posts. Alternatively we could generate them on a job every time a user favorites a post. However recommendations are generated in one big batch all together so this might be too intensive.

Only 3 recommendations are generated. In the future we could expand on this by creating a full blown recommendations page with many more recommendations, but for now this is a good start.

The recommendations use the default settings from the gem, if the recommendations end up being too inaccurate we can play with the `factors` param. The gem states it can vary wildly what is best for your project.

## User recommendations
image

## Post to post recommendations
image

### DIFF
--- a/app/views/posts/_user_recommendations.erb
+++ b/app/views/posts/_user_recommendations.erb
@@ -2,6 +2,6 @@
 
 <p class="text-small"><em>Recommendations based on your favorited codes. Favorite more codes to get better recommendations!</em></p>
 
-<main class="items">
+<section>
   <%= render collection: current_user.recommended_posts, partial: "post", as: :post %>
-</main>
+</section>
--- a/app/views/posts/_user_recommendations.erb
+++ b/app/views/posts/_user_recommendations.erb
@@ -1,6 +1,6 @@
 <h2 class="mt-1/1 mb-0"><strong>Recommended</strong> Codes</h2>
 
-<p><em>Recommendations based on your favorited codes. Favorite more codes to get better recommendations!</em></p>
+<p class="text-small"><em>Recommendations based on your favorited codes. Favorite more codes to get better recommendations!</em></p>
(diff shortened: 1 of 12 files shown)


CactusPuppy:
Looks really cool! I'll give this a review later today.

app/views/posts/_user_recommendations.erb
```diff
@@ -0,0 +1,9 @@
+<% cache [current_locale, current_user.recommended_posts] do %>
```
CactusPuppy, line 1: Does `current_user.recommended_posts` have a good equality check for the cache?

CactusPuppy (approved):
LGTM! One small comment, but other than that, I'm excited to bring this in.

app/views/posts/_user_recommendations.erb
Mitcheljager: It should do! It is just an active records collection so it should work like any other

Issue #152, opened 2022-05-18
[Feature request] Elo Hell Workshop site archive


CactusPuppy:
Elo Hell Workshops site posts need to be migrated to Workshop.codes for archival purposes.

**Current iteration TODO**:

- [x] Posts from Elo Hell site will be under an `elohell-archive` account
- [x] Original author would be indicated in title
- [x] ~~Posts from Elo Hell would include an "archived" tag~~
- [x] Backend must contain records linking archive posts to original author Battle.net UIDs
- [x] Automatic form for processing archive post deletion request via Battle.net OAuth

PR #194, opened 2022-08-30
[chore] Backend changes for #152

CactusPuppy:
This PR introduces two backend changes which will support the introduction of #152, which are:

### 1. Temporary OAuth sessions
These sessions are not backed by an actual Workshop.codes user, which allows users to request the deletion of their legacy content without needing to persist more of their own user data to Workshop.codes.

### 2. Archive authorization records
These records map import codes to the owner's Battle.net UID, which will enable Workshop.codes to know when the right user authorizes an action on a legacy post.

Tests are also included to cover the basic functionality of these two systems.


app/controllers/application_controller.rb
```diff
@@ -94,6 +95,18 @@ def render_404
 
   private
 
+  def expire_oauth_session
+    if (session[:oauth_uid].present? || session[:oauth_provider].present?) && session[:oauth_expires_at].blank?
```
Mitcheljager, line 99: Are these only set for the archive oauth?

CactusPuppy: Currently yes, but just in case we need it later, I decided make it extensible. This particular method lives here because it was the most efficient way I could think of to have any request clear the state if it was made after expiration.

config/routes.rb
```diff
@@ -131,6 +131,8 @@ def self.matches?(request)
       get ":code", to: "posts#show"
       get ":code/:tab", to: "posts#show", as: "post_tab"
       post "immortalise", to: "posts#immortalise", as: "immortalise_post"
+      patch "archive-transfer", to: "posts#archive_transfer"
```
Mitcheljager, line 134: Would it make sense to move this to a `resource` and move `archive_transfer` and `archive_destroy` to a separate controller?

Mitcheljager (changes requested):
Overall great changes! I think it would be nice to move the archive actions to their own controller. The posts controller is real big as it is already.

config/routes.rb
CactusPuppy: Conceptually I always thought of these actions as being on posts with the authorization being a supporting resource, but come to think of it doing it the other way around may make sense. I'll try it out later.

Mitcheljager (approved):
Great job!

PR #196, opened 2022-09-20
feat: Add html endpoint to wiki dictionary

Mitcheljager:
This PR adds a quick and dirty HTML endpoint to the already existing `/wiki/dictionary` json endpoint. The code isn't the nicest ever, but it's meant to be fairly rudimentary and so I'd rather not spend too much time on it.

Each item is generated from the actions and values .yaml files. Their links _should_ always match up, even though they aren't directly compared. Old article urls still redirect to new ones in case any article was renamed.

image

### DIFF
--- a/app/views/wiki/dictionary/_item.html.erb
+++ b/app/views/wiki/dictionary/_item.html.erb
@@ -1,0 +1,3 @@
+<div>
+  <%= link_to item[1]["en-US"], wiki_article_path(CGI.escape(item[1]["en-US"]).downcase) %>
+</div>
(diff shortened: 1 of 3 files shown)


app/views/wiki/dictionary/_item.html.erb
```diff
@@ -0,0 +1,3 @@
+<div>
+  <%= link_to item[1]["en-US"], wiki_article_path(CGI.escape(item[1]["en-US"]).downcase) %>
```
CactusPuppy, line 2: Is there *any* possible chance we can make this use the `I18n.current_locale` somehow?

CactusPuppy (approved):
I left one comment for a non-blocking possible improvement, but otherwise LGTM!

app/views/wiki/dictionary/_item.html.erb
Mitcheljager: That is a good idea, however I just checked and we don't have Korean translations for any of the values or actions, so we'd still only have English.

PR #199, opened 2022-09-23
feat: Add frontend for archived post actions

CactusPuppy:
This PR adds the front-end for the features introduced in #194, with the ultimate goal of closing #152.

Any archive post will have the following section after the description:
image

Upon clicking through, the user will be shown the following page if they are not logged in:
image

If the user is logged in, but unauthorized through either their main Workshop.codes account or any of their linked accounts, they will see the following page instead:
image

Upon authorizing, either automatically or manually, the user will be shown this page if they are logged in:
image
and this page if they are not:
image

Finally, if the user is unauthorized and exhausts both potential methods of authentication, they will see this page:
image
or this variant if they are logged in:
image


app/controllers/sessions_controller.rb
```diff
@@ -17,7 +17,8 @@ def create
     if should_authorize_only
       set_session_auth
       flash[:notice] = "You are now authenticated as #{auth_hash["info"]["name"] || auth_hash["info"]["battletag"]} for the next 30 minutes."
-      redirect_to "#{request.base_url}/#{omniauth_params["redirect_path"].presence || ""}"
+      path = omniauth_params["redirect_path"].presence || ""
+      redirect_to "#{request.base_url}#{path.starts_with?("/") ? "" : "/"}#{path}"
```
Mitcheljager, line 21: This is a little hard to read, it's also copied a few lines down. Would it be possible to move it to it's own method?

app/views/posts/_archive.erb
```diff
@@ -0,0 +1,7 @@
+<hr class="small" id="archive" />
```
Mitcheljager, line 1: Is this file supposed to be just `.erb` rather than `.html.erb`?

app/views/posts/show.html.erb
```diff
@@ -43,11 +43,12 @@
       <%= tabs_content_tag("description", "", "") do %>
         <%= render "carousel" if @ordered_images.present? || @post.carousel_video.present? %>
 
-        <% if @post.description.present? || @post.source_derivs.any? %>
+        <% if @post.description.present? || @post.source_derivs.any? || @archive_authorization.present? %>
           <div class="item__content item__description">
             <%= sanitized_markdown(@post.description) %>
 
             <%= render "derivatives/derivatives" if @post.source_derivs.any? %>
+            <%= render "archive", post: @post if @archive_authorization.present? %>
```
Mitcheljager, line 51: Please double check if this works fine with caching, it might cache the unauthorized version no matter what

Mitcheljager (approved):
Really nice! Left a few comments but overall I'm impressed!

app/views/archives/_archive_controls.html.erb
```diff
@@ -0,0 +1,34 @@
+<div class="form-group-columns">
+  <div class="well well--dark">
+    <p class="mt-0">You may choose to transfer this post to yourself.</p>
```
Mitcheljager, line 3: Please add headings to these columns. The headings might be the exact same as the buttons but it might make it easier to see what's going on at first glance

app/views/posts/_archive.erb
CactusPuppy: No, that's an error on my part.

app/views/posts/show.html.erb
CactusPuppy: Turns out your suspicions were correct. If an authorization is added after the post is cached, the partial is not rendered.

CactusPuppy: This issue is actually not relevant because transferring the post will update the post's `updated_at` timestamp.

CactusPuppy:
For posterity: I ultimately didn't end up adding an "archive" tag to the post card like we did with OW2 and PTR codes, mostly because I figured that slot was best reserved for when the code itself may not be valid and people copying it would need to know no matter where they were copying it from.

Mitcheljager (approved):
All looks good!

PR #200, opened 2022-09-23
chore: Overwatch 2 prep

Mitcheljager:
This PR contains a bunch of things to prep for the launch of Overwatch 2. We can merge it in now or wait until closer to release. Since there already is a bunch of OW2 related content (other maps and heroes) it might make sense to merge this anyway.

- Added Kikiro
  - Kirikos image might be temporary, I don't think we have her proper hero gallery image yet. This one was taken from a wiki
- Added Esperança
- Removed OW2 marker, filter, and routes
  - `/overwatch-2` now redirects to the homepage to still make use of the current SEO ranking of that page
- Added rake task to mark all OW2 marked posts as unlisted
- I've also removed Exclude Expired from the filter dropdown as I think it might send the wrong message
- Added banner stating the Workshop is currently disabled. This banner is behind a flag just like the OW2 banner was so it won't be shown until we want it to. The banner might be a little too intimidating for users who are only there to find codes rather than create them. Open to feedback!
image


CactusPuppy:
>     * Added banner stating the Workshop is currently disabled. This banner is behind a flag just like the OW2 banner was so it won't be shown until we want it to. The banner might be a little too intimidating for users who are only there to find codes rather than create them. Open to feedback!
>       image

Maybe we could do an orange themed version instead to catch attention but not make the warning seem "critical"?

CactusPuppy (approved):
LGTM! (Apart from one suggestion further up in this thread)

Mitcheljager:
Good thinking, looks a lot less aggressive. I hope it doesn't come across as a "Please enable your ad blocker" message, it kind of feels the same.
image

config/locales/en.yml
```diff
@@ -76,6 +76,9 @@ en:
     overwatch_2:
       title_html: <span class="text-overwatch-2">Overwatch 2</span> Beta compatible&nbsp;codes
       button: View
+    workshop_disabled:
+      title: The Workshop is currently disabled
+      content_html: With the release of Overwatch 2 the Workshop is not yet enabled. <strong class="text-white">Previously created codes can still be imported</strong>. Custom games without the Workshop are still available.
```
CactusPuppy, line 81:
```suggestion
      content_html: The Workshop is not yet enabled in Overwatch 2. <strong class="text-white">Previously created codes can still be imported</strong>. Custom games without the Workshop are still available.
```
Just took another look at the wording of the banner, and I think this version reads a little better.

PR #202, opened 2022-09-24
refactor: Move latest posts to their own action

Mitcheljager:
With the refactor of URLs pagination broke on the homepage. The homepage was always a bit messy in that it was not _just_ the homepage, it was also the latest page in disguise. With this PR latest is moved to it's own action with it's own url `/latest`. A redirect is added for SEO, as `/page/1` is a very popular page.


CactusPuppy (approved):
LGTM!

PR #207, opened 2022-11-11
Final version of workshop UI - all should work

Zezombye:


Mitcheljager:
Nice! I will make some changes and move all the files to their more rails appropriate folders, I will let you know where you can find things once it's done

PR #209, opened 2022-11-14
feat: Add ability to rename projects

CactusPuppy:
This PR adds an item to the Edit menu which allows the user to rename their project.
Rename dropdown showcase

Rename modal showcase

When the user renames their project, they will be shown a notification which informs them that the operation was successful:
Rename success notification

### DIFF
--- a/app/javascript/src/components/editor/ProjectsDropdown.svelte
+++ b/app/javascript/src/components/editor/ProjectsDropdown.svelte
@@ -240,13 +240,13 @@
 
     {#if showProjectSettings}
       <div transition:fly={{ duration: 150, y: 20 }} class="dropdown__content dropdown__content--left block w-100" style="width: 200px">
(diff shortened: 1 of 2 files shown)


app/javascript/src/components/editor/ProjectsDropdown.svelte
```diff
@@ -171,16 +209,26 @@
         </div>
       {/if}
 
-      <div class="p-1/2">
-        <h3 class="mb-0 mt-0">Create a new project</h3>
+      {#if modalType == "create"}
```
Mitcheljager, line 212: This could probably be a bit more DRY, but it's fine as is

Mitcheljager (approved):
Very nice! There are some things that could be different, but I will take care of that separately. It's definitely good enough as is. Thank you!

PR #210, opened 2022-11-14
fix: Update document title when project is renamed

CactusPuppy:
This PR fixes a minor issue where renaming a project would not update the document title

### DIFF
--- a/app/javascript/src/components/editor/Editor.svelte
+++ b/app/javascript/src/components/editor/Editor.svelte
@@ -91,6 +91,9 @@
       return string.toLowerCase().replace(/(^\w{1})|(\s+\w{1})/g, letter => letter.toUpperCase())
     }
   }
+
+  // Updates the tab title if the currentProject's title changes
+  $: document.title = $currentProject?.title !== undefined ? `Editor | ${$currentProject.title} | Workshop.codes Script Editor` : "Workshop.codes Script Editor | Workshop.codes"
 </script>
 
 <svelte:head>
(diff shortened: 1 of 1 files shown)


app/javascript/src/components/editor/Editor.svelte
```diff
@@ -91,6 +91,9 @@
       return string.toLowerCase().replace(/(^\w{1})|(\s+\w{1})/g, letter => letter.toUpperCase())
     }
   }
+
+  // Updates the tab title if the currentProject's title changes
+  $: document.title = $currentProject?.title !== undefined ? `Editor | ${ $currentProject.title } | Workshop.codes Script Editor` : "Workshop.codes Script Editor | Workshop.codes"
```
Mitcheljager, line 96: I don't think this is the correct solution. The title is set in `svelte:head` right here:
https://github.com/EloHellEsports/workshop.codes/blob/ab9f7b23704a932ea45fe04f7ab069a8e9897a05/app/javascript/src/components/editor/Editor.svelte#L96-L102

The title is most likely not updating because `$currentProject.title` is being set but that doesn't update the object.
https://github.com/EloHellEsports/workshop.codes/blob/ab9f7b23704a932ea45fe04f7ab069a8e9897a05/app/javascript/src/components/editor/ProjectsDropdown.svelte#L113
Updating an object value like this doesn't actually update the object.

So instead when renaming the project the current project could be set like this, which will update the title accordingly.
```
$currentProject = {
   ...$currentProject,
   title: value
}

Mitcheljager (changes requested):

app/javascript/src/components/editor/Editor.svelte
CactusPuppy: Upon further testing, for some reason this does not cause the document title to update. As agreed in Discord, we will instead remove the `svelte:head` component and use only the reactive statement.

PR #211, opened 2022-11-14
refactor: Ensure only one source of truth for document title

CactusPuppy:
With #210 adding another source of truth for the document title which the `svelte:head` component formerly provided, this PR eliminates the latter. The `svelte:head` component, for reasons not entirely known, does not respond to the updating of the `$currentProject` object, either by alteration of its properties or by reassignment, but the reactive statement does.


Mitcheljager (approved):
👍

PR #230, opened 2022-12-13
feat: Replace expiry message with last updated date

### LINKED ISSUE #224: Adjust how and when the Expiry message is shown
By now many many codes show they are possibly expired, but the majority of them are not. Showing the message on show pages is fine, but in the overview it might scare away users for no good reason.

CactusPuppy:
This PR replaces the expiry message (pictured below),
A listing for an example Workshop.codes post with a expiry warning
with a field showing the last revision date of the post (pictured below)
A listing for an example Workshop.codes post with the new last revision date field

KNOWN ISSUES:
- The current timeago render does not consider the user's current locale

This PR resolves #224.

### DIFF
--- a/app/javascript/src/infinite-scroll.js
+++ b/app/javascript/src/infinite-scroll.js
@@ -1,4 +1,5 @@
 import Rails from "@rails/ujs"
+import * as timeago from "timeago.js"
 
 export function bind() {
   const element = document.querySelector("[data-role='infinite-scroll-marker']")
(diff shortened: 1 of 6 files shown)


app/javascript/src/infinite-scroll.js
```diff
@@ -66,6 +67,14 @@ function getInfiniteScrollContent(element) {
         element.innerHTML = "Load more"
         element.setAttribute("data-url", requestUrlString)
       }
+
+      // Initialize "Last updated" attribute in newly loaded items
+      const elements = document.querySelectorAll("[data-role~='timeago']")
```
Mitcheljager, line 72: Perhaps instead of doing this here again, you could import our timeago.js file and call `render()` again (might need to import it as a nicer name)

Mitcheljager (changes requested):

Mitcheljager (approved):
Very nice!

PR #238, opened 2023-01-13
refactor/fix: Prevent unintended styling under `.standout` elements

CactusPuppy:
This PR fixes an issue where certain parts of the site which use the `standout` CSS class had unintentional/unintuitive styling applied. The issue arose due to styling initially added to highlight Workshop alternatives on the front page targeting generic elements instead of elements with certain classes. This PR changes the styling for the front page to use specific classes.

Example of broken styling:
image

Inspector showing the specificity override from frontpage styling:

Fixed styling:

Inspector showing lack of styling pollution:
image

### DIFF
--- a/app/assets/stylesheets/elements/_standout.scss
+++ b/app/assets/stylesheets/elements/_standout.scss
@@ -88,25 +88,25 @@
       background: var(--standout-color, #{$overwatch-2});
     }
   }
+}
 
-  &__subtitle {
-    color: $text-color-dark;
-  }
+.standout__subtitle {
+  color: $text-color-dark;
+}
 
-  &__heading {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/elements/_standout.scss
```diff
@@ -89,15 +89,15 @@
     }
   }
 
-  small {
+  &__subtitle {
```
Mitcheljager, line 92: While this syntax works, it makes it very difficult to search-all for when debugging. Preferably it would be it's own full class.

Mitcheljager (changes requested):
Good catch and overall good improvement. But there are some (all the same) small changes I'd like to see

Mitcheljager (approved):

PR #245, opened 2023-02-04
feat: Editor dynamic translations

Mitcheljager:
This PR includes an interface to add dynamic translations to your code. A UI is provided to select the languages you wish to add translations for and a UI to add different keys with textareas for each selected language. The keys can be included similarly to mixins, but by using `@translate("Key")`. If a translation is missing the Default language will be used instead, this can be set to any of the languages.

A key will be replaced with a bit of Workshop code that looks like this:
```
Value In Array(
      Array(Custom String("Something in english", Icon String(Arrow: Down)), Custom String("Something in german", Icon String(Arrow: Down)), Custom String("Something in Spanish", Icon String(Arrow: Down)), Custom String("Something in english", Icon String(Arrow: Down))),
      Max(False, Index Of Array Value(Global.WCDynamicLanguages, Custom String("{0}", Map(Practice Range), Null, Null)))
```
Additionally a rule is inserted at the bottom of your code that looks like this:
```
rule("Workshop.codes Editor Dynamic Language - Set Languages") {
    event { Ongoing - Global; }
    actions { Global.WCDynamicLanguages = Array(Custom String("Practice Range"), Custom String("Trainingsbereich"), Custom String("Práctica de combate"), Custom String("Obszar treningowy")); }
  }
```
Together these check the name of the Practice Range _client side_, meaning languages are dynamic and not set by the creator of the code. More info here: https://github.com/Zezombye/overpy/wiki/Dynamic-translations

### Button in top nav
image

### Select languages
image

### Add translations for each language
image

### Keys need to be unique but are case sensitive
image

### Keys can be renamed and removed after selected this key
image

### Key inserted in place of custom string
image

### DIFF
--- a/app/javascript/src/components/editor/Empty.svelte
+++ b/app/javascript/src/components/editor/Empty.svelte
@@ -16,6 +16,7 @@
         <li>Integrated <mark class="text-white">Wiki</mark> search.</li>
         <li>Automatic <mark class="text-white">error detection</mark> (it's a bit limited, but still).</li>
         <li>Reduce code repetition using <mark class="text-white">Mixins</mark>.</li>
+        <li>Each to manage <mark class="text-white">translation keys</mark> for strings.</li>
       </ul>
     <p>
 
@@ -30,9 +31,7 @@
       <ul>
         <li>Automatic wiki searching on auto complete.</li>
         <li>Optional semicolons.</li>
-        <li>Search and replace across all files.</li>
         <li>More customisation; colors, fonts, sizes, etc.</li>
-        <li>OverPy! Writing scripts using a HLL is much more enjoyable, getting OverPy in is a high priority.</li>
       </ul>
     <p>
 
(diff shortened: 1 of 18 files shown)


app/javascript/src/components/editor/Empty.svelte
```diff
@@ -42,7 +41,6 @@
         <li>Variables that are only stated in a For Global (or Player) Variable Loop will not be properly compiled.</li>
         <li>Some actions or values may incorrectly indicate a wrong number of arguments.</li>
         <li>Rules without names might not be properly imported.</li>
-        <li>You can't rename projects.</li>
```
CactusPuppy, line 45: You ever just add a feature and then forget to tell people you added the feature? Couldn't be me.

Mitcheljager: :D

PR #257, opened 2023-02-21
chore: Rebrand away from Elo Hell

Mitcheljager:
Elo Hell is no more, and with that neither is the branding. This PR removes all Elo Hell branding and goes back to our own branding instead. For this I've created a logo (subject to change) and simply used the OW2 color as brand color. The Editor will now use Wiki colors to avoid confusion with the orange color perhaps signaling warnings.

Some highlights:
- New logo, also used in og-images
- One of the fonts was removed, this was mainly used for titles
- Adjusted discord links on rules and wiki pages

## Home
image

## Posts
image

## Wiki
image

## Editor
image

## Footer
image


CactusPuppy (approved):
LGTM!

PR #259, opened 2023-03-04
Add links to markdown-generated headers

netux:
Headers rendered with `ContentHelper:markdown` now get an id derived from the content of the header, and a nested link to change the hash of the page to that id, so users can quickly jump to sections of a markdown document.

I checked on the default renderer from Redcarpet. While it does have a similar option, it is only active when using Table of Contents, which we don't always want.

As part of this change, I had add a style rule and used :is() for simplicity. :is() seems to have [good browser support](https://caniuse.com/?search=%3Ais), though I'm not sure which browser we want to support here.

### DIFF
--- a/app/assets/stylesheets/general/_titles.scss
+++ b/app/assets/stylesheets/general/_titles.scss
@@ -47,7 +47,12 @@ h6 {
   font-size: $font-size-base;
 }
 
-:is(h1, h2, h3, h4, h5, h6) .header_anchor {
+h1 .header_anchor,
+h2 .header_anchor,
+h3 .header_anchor,
+h4 .header_anchor,
+h5 .header_anchor,
+h6 .header_anchor {
   text-decoration: none;
 }
 
--- a/app/assets/stylesheets/general/_titles.scss
+++ b/app/assets/stylesheets/general/_titles.scss
@@ -47,15 +47,11 @@ h6 {
   font-size: $font-size-base;
 }
 
-h1 .header_anchor,
-h2 .header_anchor,
-h3 .header_anchor,
-h4 .header_anchor,
-h5 .header_anchor,
-h6 .header_anchor {
-  text-decoration: none;
+h1, h2, h3, h4, h5, h6 {
+  .header_anchor {
+    text-decoration: none;
+  }
 }
-
 .title-icon {
   display: inline-block;
   height: .75em;
--- a/app/assets/stylesheets/general/_titles.scss
+++ b/app/assets/stylesheets/general/_titles.scss
@@ -47,6 +47,10 @@ h6 {
   font-size: $font-size-base;
 }
 
+:is(h1, h2, h3, h4, h5, h6) .header_anchor {
+  text-decoration: none;
+}
+
 .title-icon {
   display: inline-block;
   height: .75em;
--- a/app/helpers/content_helper.rb
+++ b/app/helpers/content_helper.rb
@@ -44,18 +44,23 @@ module ContentHelper
       tag = "h#{ level }"
       hash = header_anchor_hash(title)
 
-      "<#{ tag } id=\"#{ hash }\"><a class=\"header_anchor\" href=\"\##{ hash }\">#{ title }</a></#{ tag }>"
+      if @options[:header_anchors]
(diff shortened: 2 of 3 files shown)


app/assets/stylesheets/general/_titles.scss
```diff
@@ -47,6 +47,10 @@ h6 {
   font-size: $font-size-base;
 }
 
+:is(h1, h2, h3, h4, h5, h6) .header_anchor {
```
Mitcheljager, line 50: Unfortunately this isn't supported by safari 13 which is still used by a significant enough portion of users to need support for it. But luckily we are using SCSS so instead you can nest it like

```
h1, h2, h3... {
   .header-anchor
}
```

Perhaps just targeting the link tag instead of giving it an explicit class name would be enough too `a[href^="#"]`, although it doesn't matter too much

app/helpers/content_helper.rb
```diff
@@ -9,6 +9,43 @@ def block_code(code, language)
...
+        .gsub(/[^a-z0-9\- ]/i, "")
+        .strip
+        .gsub(/ +/, "-")
+    end
+
+    def header(title, level)
+      tag = "h#{ level }"
+      hash = header_anchor_hash(title)
+
+      "<#{ tag } id=\"#{ hash }\"><a class=\"header_anchor\" href=\"\##{ hash }\">#{ title }</#{ tag }>"
```
Mitcheljager, line 47: Should the `a` tag not be closed as well at some point?

Mitcheljager (changes requested):
Great initiative! I'm not sure how I feel about the change overall, if this was just for the wiki I'd merge it in because that seems useful, but it also affects posts and blog articles. Which isn't as great for a couple reasons;

- Turning every header into a link isn't great ux.
- I haven't checked but I'm not sure how this interacts places where the markdown is embedded, like inside of the editor wiki sidebar. Does it still work as expected there?
- Using the anchors inside of your own post description isn't great, as there is no way to get the link inside of the markdown editor without going into the code. Currently anchors can be inserted manually with something like `<div id="some-anchor" />`. Which also requires some knowledge most people won't know, but it's easier to use if you know it's there.

netux:
Fixed both comments in the commits above.

As for the UX, I thought about it, but I wanted this to be quick thing and didn't feel like finding a link icon to be honest.
I saw the [VSCode changelog page](https://code.visualstudio.com/updates/v1_76#_fixedwidth-centered-layout) did it this way, so I thought it was fine.
It's a somewhat power-user feature anyway. If people find this by accident I don't think it's that bad. They get some feedback that the headers are links from putting their mouse over it and the cursor becoming a pointer.

app/assets/stylesheets/general/_titles.scss
```diff
@@ -47,6 +47,15 @@ h6 {
...
 
+h1 .header_anchor,
+h2 .header_anchor,
+h3 .header_anchor,
+h4 .header_anchor,
+h5 .header_anchor,
+h6 .header_anchor {
+  text-decoration: none;
+}
+
```
CactusPuppy, line 58:
```suggestion
h1, h2, h3, h4, h5, h6 {
  .header_anchor {
    text-decoration: none;
  }
}
```
Consider utilizing the ability of SCSS to nest selection levels to avoid repetition while achieving the same thing.

app/helpers/content_helper.rb
```diff
@@ -9,16 +9,58 @@ def block_code(code, language)
     def image(link, title, alt_text)
       image_tag(link, title: title, alt: alt_text, loading: "lazy")
     end
+
+    # losely based on https://github.com/vmg/redcarpet/blob/3e3f0b522fbe9283ba450334b5cec7a439dc0955/ext/redcarpet/html.c#L297
```
CactusPuppy, line 13:
```suggestion
    # loosely based on https://github.com/vmg/redcarpet/blob/3e3f0b522fbe9283ba450334b5cec7a439dc0955/ext/redcarpet/html.c#L297
```
Minor spelling mistake

CactusPuppy (changes requested):
Overall great changes! Left a few suggestions on things to tidy up before merging.

CactusPuppy (approved):
LGTM

PR #260, opened 2023-03-04
Add files for a VSCode + asdf setup

netux:
As part of setting this up in my environment, I had to do the following:

1. Add .tools-versions file so asdf can link the correct versions of ruby and node on this folder
2. Add .vscode/ folder. This wasn't strictly necessary but I'm sure some people may find this useful.
    - Added extensions.json with some recommended extensions
    - Added a quick task to kill the server (I wound up needing this a few times 😅)
3. Add execute flag to all files under bin/


CactusPuppy (changes requested):
Why are the binaries under `bin/` altered? Please remove the alteration of the `bin/` folder and ignore the changes to the binaries if the changes are required to run the repository locally.

.vscode/tasks.json
```diff
@@ -0,0 +1,12 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=733558
+	// for the documentation about the tasks.json format
+	"version": "2.0.0",
+	"tasks": [
+		{
+			"label": "Kill server",
+			"type": "shell",
+			"command": "kill --KILL (cat \"${workspaceFolder}/tmp/pids/server.pid\") && rm \"${workspaceFolder}/tmp/pids/server.pid\""
+		}
```
CactusPuppy, line 10: Is there a situation where a simple `Ctrl + C` doesn't kill the server?

Mitcheljager: Sometimes pids remains after unexpected closings. Happens to me sometimes when restarting my PC while the server is still on. I always fix it by manually removing the pids folder.

netux: > Is there a situation where a simple Ctrl + C doesn't kill the server?

I accidentally create an infinite loop while serving a request earlier when working on #259. Ctrl+C is handled, but the program refuses to shut down until all requests are handled. Because of the infinite loop, that would never be the case.

Maybe this was a very specific case though. Can't really think of other ones.

netux:
The files under `bin/` were modified for the same reason as discussed on Discord the other day: I had to make them executable so I could run the webpack dev server. As this PR is more of a "here are the changes for my setup to work", I decided to include these here as well.

I know you said I shouldn't commit this, but I'd like to argue it makes sense for people in the future cloning the repo from scratch. Other people running this on Unix/Linux will probably have the same problem as me.

I just digged a little more info about this btw. Turns out I have [`core.filemode` on my git config enabled](https://stackoverflow.com/a/1257613/7492433) (I believe it's enabled by default?), so that's why the file mode change were staged.
My change here changes the file mode from [644](https://chmodcommand.com/chmod-644/) (owner readable + writeable, group readable, others readable) to [755](https://chmodcommand.com/chmod-755/) (owner rwx, group readable + executable, others readable + executable). You can also see that on GitHub:
image

CactusPuppy:
> The files under `bin/` were modified for the same reason as discussed on Discord the other day: I had to make them executable so I could run the webpack dev server. As this PR is more of a "here are the changes for my setup to work", I decided to include these here as well.
> 
> I know you said I shouldn't commit this, but I'd like to argue it makes sense for people in the future cloning the repo from scratch. Other people running this on Unix/Linux will probably have the same problem as me.
> 
> I just digged a little more info about this btw. Turns out I have [`core.filemode` on my git config enabled](https://stackoverflow.com/a/1257613/7492433) (I believe it's enabled by default?), so that's why the file mode change were staged. My change here changes the file mode from [644](https://chmodcommand.com/chmod-644/) (owner readable + writeable, group readable, others readable) to [755](https://chmodcommand.com/chmod-755/) (owner rwx, group readable + executable, others readable + executable). You can also see that on GitHub: image

As eventually agreed upon in the Discord, executable permissions on the scripts in `bin/` is better to make onboarding of newer contributors easier.

CactusPuppy (approved):
LGTM

PR #261, opened 2023-03-05
Move markdown header anchor next to header, represent with a chain icon

netux:
Following up with comments on #259, I decided to work on the improved UX anyway.

In this PR, instead of having the header be the link to itself, we move the link to be a small chain/link icon next to the header. This icon only appears when the user hovers over the header.

This is more inline with how places like GitHub do this.

Credits for the chain icon are inside the file. I just picked one semi-randomly. This one seems to come from a Yandex UI library.


app/assets/stylesheets/general/_titles.scss
```diff
@@ -49,7 +49,21 @@ h6 {
 
 h1, h2, h3, h4, h5, h6 {
   .header_anchor {
-    text-decoration: none;
+    float: left;
+    margin-left: -20px;
+    width: 20px;
+    height: 1em;
+    line-height: 1;
+
```
Mitcheljager, line 57: No new lines in between css properties please. I usually order css by `position` `block model` `font`, `rest`, but feel free to ignore for now.

app/assets/stylesheets/general/_titles.scss
```diff
@@ -49,7 +49,21 @@ h6 {
 
 h1, h2, h3, h4, h5, h6 {
   .header_anchor {
-    text-decoration: none;
+    float: left;
```
Mitcheljager, line 52: We shouldn't use floats as they tend to break layouts. `position: absolute` might be more suitable here.

app/helpers/content_helper.rb
```diff
@@ -45,7 +45,10 @@ def header(title, level)
       hash = header_anchor_hash(title)
 
       if @options[:header_anchors]
-        "<#{ tag } id=\"#{ hash }\"><a class=\"header_anchor\" href=\"\##{ hash }\">#{ title }</a></#{ tag }>"
+        "<#{ tag } id=\"#{ hash }\">" \
```
Mitcheljager, line 48: I think in other places we don't escape the double quotes, we use single quotes instead. Similarly with the \ new line, I think it's nicer if it's just one big string.

app/assets/stylesheets/general/_titles.scss
```diff
@@ -49,7 +49,21 @@ h6 {
...
+    margin-left: -20px;
+    width: 20px;
+    height: 1em;
+    line-height: 1;
+
+    background: transparent image-url("icons/icon-link.svg") no-repeat;
+    background-position: left center;
+
+    opacity: 0;
```
Mitcheljager, line 61: I think we might need to use `display: none` to hide the button, otherwise you might accidentally hit it with touch controls

app/assets/stylesheets/general/_titles.scss
```diff
@@ -49,7 +49,21 @@ h6 {
 
 h1, h2, h3, h4, h5, h6 {
   .header_anchor {
```
Mitcheljager, line 51: I didn't spot this before, but we use `kebab-case` for regular class names

Mitcheljager (changes requested):
Left a couple comments, all code formatting, sorry about that :D. Other than that it seems like a great change

app/assets/stylesheets/general/_titles.scss
```diff
@@ -49,7 +49,21 @@ h6 {
...
-    text-decoration: none;
+    float: left;
+    margin-left: -20px;
+    width: 20px;
+    height: 1em;
+    line-height: 1;
```
Mitcheljager, line 56: I feel like the line-height does nothing here

app/assets/stylesheets/general/_titles.scss
netux: I took this from the way GitHub was doing it. `position: absolute` means the headers need `position: relative`, which is something I wanted to avoid. `float: left` seems to work for allowing `width` and `height` to change.

app/helpers/content_helper.rb
netux: For the new lines, what would you suggest? I didn't want it to be a big single line, nor have extra spacing on the output.

app/assets/stylesheets/general/_titles.scss
netux: Also stolen from GitHub. Indeed it doesn't do anything. I will remove 👍.

app/helpers/content_helper.rb
Mitcheljager: Here for example it's just a single string. What the output looks like doesn't matter as it all gets minified anyway.

https://github.com/Mitcheljager/workshop.codes/blob/b3c2db9db11f40a0e89c785c1d260f9582c24026/app/helpers/content_helper.rb#L85-L87

app/assets/stylesheets/general/_titles.scss
CactusPuppy: @netux Any particular reason you don't want to use `position: absolute`?

netux: > `position: absolute` means the headers need `position: relative`

I'd argue that changes the layout even more than `float`. And I see no reason to change it when this works fine too.

Any particular examples of `float` breaking layout you may like to point out? Again, I'm mostly copying what GitHub does so I assume they didn't find any issues with it.

CactusPuppy: `position: relative` doesn't change anything about the layout outside of its children.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/position#syntax):
> `relative`
>
>    The element is positioned according to the normal flow of the document, and then offset relative to itself based on the values of top, right, bottom, and left. The offset does not affect the position of any other elements; thus, the space given for the element in the page layout is the same as if position were static.

Since headings are block-level elements anyways, nothing outside of the heading's own context would be affected.

CactusPuppy (approved):
LGTM

PR #262, opened 2023-03-05
Convert search result items on editor wiki search from buttons to links

netux:
While the inline wiki is pretty useful, sometimes you want to open multiple browser tabs and have different wikis open, and the search can help with that.

With this change, search result items now behave like normal `<a>`s, meaning you can middle click to open the article on a new tab, or use the browser's context menu for links.

### DIFF
--- a/app/javascript/src/components/editor/EditorWikiSearch.svelte
+++ b/app/javascript/src/components/editor/EditorWikiSearch.svelte
@@ -67,10 +67,7 @@
           class="editor-wiki-results__item"
           href={`/wiki/articles/${result.slug}`}
           target="_blank"
-          on:click={(event) => {
-            event.preventDefault();
-            selectResult(result);
-          }}
+          on:click|preventDefault={() => selectResult(result)}
         >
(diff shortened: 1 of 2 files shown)


app/javascript/src/components/editor/EditorWikiSearch.svelte
```diff
@@ -63,10 +63,18 @@
   {#if loading || results.length}
     <div class="editor-wiki-results">
       {#each results as result}
-        <button class="editor-wiki-results__item" on:click={() => selectResult(result)}>
+        <a
+          class="editor-wiki-results__item"
+          href={`/wiki/articles/${result.slug}`}
+          target="_blank"
+          on:click={(event) => {
+            event.preventDefault();
```
Mitcheljager, line 71: Instead of event.preventDefault you can do `on:click|preventDefault=...`

Mitcheljager (changes requested):
Great change, small suggestion but nothing blocking. Nice one!

PR #263, opened 2023-03-05
feat: Atomically update post and derivatives

CactusPuppy:
This PR creates a better UX for situations where a derivative may be invalid (e.g. when a derivative has the same import code as the post itself). Previously, the post creation would partially succeed, but the invalid record would cause an invisible error, which would potentially fail future attemps to save.

Now, if any part of the post form causes an error, the entire operation will atomically fail, meaning that a partial success save is no longer possible. In addition, a proper error is returned identifying which derivation caused the issue and for what reason.

image

### DIFF
--- a/app/assets/stylesheets/general/_form.scss
+++ b/app/assets/stylesheets/general/_form.scss
@@ -268,8 +268,13 @@ input[type="file"] {
   display: inline;
 
   input,
-  select {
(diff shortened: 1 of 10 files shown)


Mitcheljager (approved):
Makes sense, very nice!

PR #264, opened 2023-03-07
feat: Auto-fill search input when pressing Ctrl + Shift + F

CactusPuppy:
Demonstration of PR in action:
firefox_cQiN33b8wG

### DIFF
--- a/app/javascript/src/components/editor/FindReplaceAll.svelte
+++ b/app/javascript/src/components/editor/FindReplaceAll.svelte
@@ -1,5 +1,5 @@
 <script>
-  import { items } from "../../stores/editor"
+  import { items, currentItem, editorStates } from "../../stores/editor"
   import { getItemById, replaceBetween, setCurrentItemById, updateItem } from "../../utils/editor"
   import { fade, fly } from "svelte/transition"
   import { tick } from "svelte"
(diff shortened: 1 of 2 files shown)


CactusPuppy:
For some reason, the `editorStates` store doesn't properly track when the user selects something.

app/javascript/src/components/editor/FindReplaceAll.svelte
```diff
@@ -115,7 +115,11 @@
     if (event.ctrlKey && event.shiftKey && event.keyCode == 70) { // F key
       event.preventDefault()
       active = !active
-      if (active) focusInput()
+      if (active) {
+        focusInput()
+        const codeMirrorSelectedText = $editorStates[$currentItem.id].sliceDoc($editorStates[$currentItem.id].selection.main.from, $editorStates[$currentItem.id].selection.main.to)
```
Mitcheljager, line 120: Perhaps `$editorStates[$currentItem.id].selection.main.from, $editorStates[$currentItem.id].selection.main.to` could be more readable like
```suggestion
const [from, to] = $editorStates[$currentItem.id].selection.main
const codeMirrorSelectedText = $editorStates[$currentItem.id].sliceDoc(from, to)
```

Mitcheljager (changes requested):
Very nice overall, nice and simple. One small suggestion, but it's also good as is

app/javascript/src/components/editor/FindReplaceAll.svelte
CactusPuppy: Oooh, I like it!

Issue #250, opened 2023-02-04
[Editor] Auto indenting is often wrong


Mitcheljager:
When moving something to a new line using Enter the auto indenting is often wrong. I'm not sure of the exact cause or situation this happens in, but often times when trying to align each parameter in an action on it's own line will be auto indenting place it one intend too far.

It's a bit vague and not reliable to reproduce, but it happens often.

CactusPuppy:
Probably fixed in #268

PR #268, opened 2023-03-09
Fix for Auto indenting often being wrong #250

Loottoo:
the regex for const spaces matches both spaces and tabs.  There might be a better regex to grab only spaces without catching the tabs, but an easy fix is to just subtract the number of tabs instead.


Mitcheljager (approved):
Nice catch!

PR #270, opened 2023-03-13
Change Preview button to "Edit" when previewing in the markdown editor

netux:
This is a hack on top of a hack to get the Preview button to say "Preview" in the first place.

Instead of simply setting the ::before content of `.fd-eye` to "Edit", I make the ::before invisible and use "Edit" on the ::after content. This is so the button preserves its width, and the other buttons don't shift around.

preview → edit


Mitcheljager (approved):
Hack indeed, but sure, that works!

PR #271, opened 2023-03-13
Direct verified users to their profile page to edit custom css

netux:
I actually spent 15 minutes looking for it... with the code for it next to me 🫠 

Wording could be better here. I'll leave that open to suggestions from you.


CactusPuppy (changes requested):
I like the idea of making the Custom CSS easier to find! Maybe putting "Customize your profile" blurb within the "You're verified!" block on the "About You" tab would be nicer though.

PR #272, opened 2023-03-13
Set custom CSS with `.textContent` instead of `.innerHTML`

netux:
Even if you have a sanitizer in the middle, sanitizers tend to be bypassable.


Mitcheljager:
>Even if you have a sanitizer in the middle, sanitizers tend to be bypassable.

I don't think it is that easily bypassable, but it can happen of course. But even if so, this change doesn't mitigate that as this javascript only runs while editing the css. The final css that's rendered on the page is done directly on the server.

CactusPuppy:
This code is currently only applied to input provided by the user, with no danger of any kind of one-click attacks. That being said, there's no harm in using best practices.

PR #275, opened 2023-03-17
style: Safari pass

Mitcheljager:
This PR fixes a couple issues in Safari, specifically for iOS

- The hamburger menu now uses an svg rather than a text icon as some default fonts handle them poorly
- Mobile navigation icons now have a fixed width to make it more predictable
- Text in snippets has a set font-size-adjust to overwrites Safaris good intentions
- Thumbnails in posts and cards will now use higher resolutions on high pixel density displays


CactusPuppy (approved):
LGTM

PR #276, opened 2023-03-25
feat: Wiki quick navigation

Mitcheljager:
To make it a little easier to scroll through the wiki this PR introduced the "quick navigation". This is a list of all Workshop keywords separated by their categories that is visible on all article pages.

image

- The navigation consists of all Workshop values that are also used by the editor. It doesn't use the actual articles to avoid having to make a giant query. The lists are sorted alphabetically

- It's only visible on large enough screens. On smaller screens there's no alternative, it's just not there at all. Perhaps at some point we can think of a way to also do this for mobile screens, but for now just desktop will do.

- This navigation adjusts to the height of your screen and the height of the container so that it always stays scrollable within your viewport.
wiki-nav

- When visiting a page the category of the current article is automatically expanded and if a corresponding link is found it is highlighted and scrolled into view.
image

- When you are browsing an article that is not within any of the categories all of the categories are collapsed
image

- The scrollbar is only visible when hovering the box

| Firefox | Chrome |
| --- | --- |
| scroll-firefox | scroll-chrome |


app/views/wiki/_quick_navigation.html.erb
```diff
@@ -0,0 +1,46 @@
...
+
+          <div class="quick-navigation__icon">&#8250;</div>
+        </button>
+
+        <div data-role="content-to-toggle" style="<%= "display: none" if @article.category.title != category.to_s %>">
+          <% props.each do |content| %>
+            <%= link_to content["en-US"],
+              CGI.escape(content["en-US"]).downcase,
+              class: "quick-navigation__item #{ "quick-navigation__item--active" if @article.title == content["en-US"] }",
+              data: { prefetch: false, role: ( 'scroll-into-view-on-load' if @article.title == content["en-US"]) } %>
```
CactusPuppy, line 23: It might be nice to make the active item not a link at all to avoid allowing users to reload the page by accidentally clicking on it.

CactusPuppy (changes requested):
Overall, great idea! Have a few nitpicks for appearance and minor functionality, but once those are solved, LGTM!

app/views/wiki/_quick_navigation.html.erb
Mitcheljager: I see what you mean, but I'd rather keep it as is, even if only to make it not unnecessarily complex.

app/javascript/src/scroll-into-view-on-load.js
```diff
@@ -0,0 +1,9 @@
+export function initialize() {
+  const elements = document.querySelectorAll("[data-role~='scroll-into-view-on-load']")
+
+  elements.forEach(element => {
+    const parent = element.closest("[data-scroll-container]")
+
+    parent.scrollTop = element.offsetTop
```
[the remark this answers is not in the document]
Mitcheljager: I solved it a little different, by subtracting a set value. Not all items are the same height, some are 2 (or even 3) lines high which results in a much larger offset.

app/views/wiki/_quick_navigation.html.erb
CactusPuppy: Fair enough

CactusPuppy (approved):
LGTM!

PR #277, opened 2023-04-01
Add upper Y bound to April Fools 2023 Impetus

netux:
It was possible for the element to go too far up, and not be recoverable. This PR adds an upper bound (but leaves the lower bound at Infinity) so we don't have to worry about that.

### DIFF
--- a/app/javascript/src/april-fools.js
+++ b/app/javascript/src/april-fools.js
@@ -25,6 +25,7 @@ async function hiimpetus(event) {
     source: element,
     friction: 0.96,
     boundX: [(window.innerWidth - element.offsetWidth - 100) * -0.5, (window.innerWidth - element.offsetWidth - 100) * 0.5],
+    boundY: [-element.offsetHeight, Infinity],
     update: function(x, y) {
       element.style.transform = `translateX(${ x }px) translateY(${ y }px)`
       element.dataset.x = x


app/javascript/src/april-fools.js
```diff
@@ -25,6 +25,7 @@ async function hiimpetus(event) {
     source: element,
     friction: 0.96,
     boundX: [(window.innerWidth - element.offsetWidth - 100) * -0.5, (window.innerWidth - element.offsetWidth - 100) * 0.5],
+    boundY: [-element.offsetHeight, Infinity],
```
Mitcheljager, line 28: Is this correct? Won't this cause it to not be able to go up all that far but still go off document down?

netux: Yes, that was my intention. If it goes below the document the scrollbar expands, so you can still scroll down and grab it again.

I could've added a lower bound as well, but I didn't feel like figuring out how to calculate the page's height with scrolling.

Mitcheljager: I like the intention, but I think I'm ok with them going offscreen. Throwing them around aimlessly is fun!

netux: But then it's inconsistent: it can go out of the screen vertically, but bounces horizontally...
Isn't it more fun to not lose the damn thing if it goes off screen? That way you can keep playing with it 😜

CactusPuppy:
Closing as the April Fools' event is now over.

PR #278, opened 2023-04-03
adding missing colors and events

Loottoo:
- Added missing colors and events.
- Fixed a bug where Color would autofill Color(Color(White)).
- Changed default number to 0 (making for example vector default Vector(0, 0, 0) rather than Vector(1, 1, 1). for loops also begin at index 0, rather than 1).

### DIFF
--- a/app/assets/stylesheets/elements/_codemirror.scss
+++ b/app/assets/stylesheets/elements/_codemirror.scss
@@ -114,7 +114,9 @@
   .cm-completionIcon {
     padding-right: 1rem;
   }
-
+  .cm-completionIcon-event::after {
+    content: "∈";
+  }
   .cm-tooltip {
     background: darken($bg-dark, 5%);
     border: 0;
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/elements/_codemirror.scss
```diff
@@ -114,7 +114,9 @@
   .cm-completionIcon {
     padding-right: 1rem;
   }
-
+  .cm-completionIcon-event::after {
+    content: "∈";
+  }
```
CactusPuppy, line 119: What's this for?

config/arrays/wiki/defaults.yml
```diff
@@ -2,12 +2,12 @@ Compare: "True"
 Player: Event Player
 Primary Fire: Button(Primary Fire)
 Vector: Vector(0, 0, 0)
-Number: 1
+Number: 0
+Number (1): 1
```
CactusPuppy, line 6: Kind of ugly, but it works.

app/assets/stylesheets/elements/_codemirror.scss
Loottoo: its the symbol shown in the editor, E for Event!
image

config/arrays/wiki/defaults.yml
Loottoo: yeah, some values really should default to 0, like the index for loops. others really shouldnt, like the incremented number or chase rate. not sure what else would be a better solution

PR #279, opened 2023-04-06
Fix for subroutines offsetting the linefinder

Loottoo:
Linefinder would count too many lines after regex matched the lines with linemarkers. removing the double check for subroutines ensures there wont be any duplicates. all subroutines should still be found assuming any called subroutine actually exists.


Mitcheljager (approved):
Nice find!

PR #280, opened 2023-04-11
Remove z-index property from DragHandle in Editor

netux:
This made the handle go over dropdowns like the project select.


Mitcheljager (approved):
Nice, seems to work fine still so it's probably a left over from during development

PR #283, opened 2023-04-12
Add simple VSCode Chrome launch task

netux:
The launch task only opens the browser on localhost:3000. It does not start the webpack nor rails server.

The benefit of this is that the launched profile is persistant and isolated from your browser profiles. So you can install extensions specific to your workflow (e.g. Svelte DevTools, Rails Panel) and they won't clutter your normal browser experience.

---

This is a pattern I've been using recently, but I understand if it a little niche (even more than #260).

It might be worth to also include launch tasks for webpack and rails, but so far I'm okay with running the commands on the terminal and I didn't feel like debugging launch tasks.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,15 @@
+{
+	// Use IntelliSense to learn about possible attributes.
+	// Hover to view descriptions of existing attributes.
+	// For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+	"version": "0.2.0",
+	"configurations": [
+		{
+			"type": "chrome",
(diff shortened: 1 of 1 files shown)


netux:
> I have no idea what this means, can you explain what's going on?

This is just a shortcut to open Chrome with a custom profile for the project, so no bookmarks/extensions/local data shared with the default profile you use.

Open VSCode → Start webpack and rails server → Hit F5 → Browser opens in localhost:3000:
image

Mitcheljager:
Well sure! It's not something I would personally use, but might as well include it I suppose

PR #284, opened 2023-04-12
Make Editor Workshop.codes logo take you to the home page, add icon to still go to the Editor Welcome page

netux:
- The workshop.codes logo now takes you to https://workshop.codes/
- The new Info icon keeps the old behavior of the logo
image

I'm not too happy with the info icon (I quickly took it from fontawesome.com, suggestions are welcome here).

---

Also: make the new icon that takes you to the Editor Welcome page replace the URL (`/editor?uuid=...` → `/editor`). This is done in a way that users can use the browser's back navigation button to take them back to the project you were working on. See commits for details.

### DIFF
--- a/app/javascript/src/components/editor/Editor.svelte
+++ b/app/javascript/src/components/editor/Editor.svelte
@@ -17,8 +17,8 @@
   import ItemFinder from "./ItemFinder.svelte"
   import FindReplaceAll from "./FindReplaceAll.svelte"
   import LineFinder from "./LineFinder.svelte"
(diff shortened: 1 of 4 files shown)


CactusPuppy:
I don't love the information icon, because usually those indicate a helpful pop-up, not a full page navigation.

Mitcheljager:
I think I will have to reject this change. It's a clever enough solution, but I think it makes sense for the Editor to be in it's own space with it's own context if that makes sense.

PR #288, opened 2023-04-13
feat: Duplicate Project

### LINKED ISSUE #285: [Editor] "Duplicate Project" button
Users have expressed the desire to be able to duplicate an existing project for the purposes of making variations of the same project.

Mitcheljager:
Closes #285 

This PR adds the ability to duplicate a project, creating an exact copy of the content, but of course as a completely new project. This is done by simply passing the content along when creating a project. 

But you may be looking at this thinking "Gee, that's a lot of changes just for duplicating", and you'd be right. There was too much going on in the ProjectsDropdown so I decided to split it up.
1. First up, all common fetch functions have been moved to `utils/project.js`. This includes creating, updating, and destroying a project, among more things.
2. Next the Create modal was moved to it's own component with it's own logic, rather than being housed directly in the ProjectsDropdown component. The modal is still a little bit weird in that it also uses logic from non svelte components from our standard modal javascript. I will leave that refactor for another day. For now it was about clearing up the ProjectsDropdown component.

## Duplicating
duplicate-project

### DIFF
--- a/app/javascript/src/components/editor/Modals/CreateProjectModal.svelte
+++ b/app/javascript/src/components/editor/Modals/CreateProjectModal.svelte
@@ -1,0 +1,94 @@
+<script>
+  import { currentProject, isSignedIn } from "../../../stores/editor"
+  import { createProject, renameCurrentProject } from "../../../utils/project"
+  import { createEventDispatcher } from "svelte"
+  import { fade, fly } from "svelte/transition"
+
+  const dispatch = createEventDispatcher()
+
+  let loading
+  let value
+  let active = false
+  let modalType = "create"
+
+  export function showModalOfType(type, title = "") {
+    modalType = type
+    value = title
+    active = true
+  }
+
+  async function newProject() {
+    loading = true
+
+    const data = await createProject(value)
+    if (data) {
+      dispatch("setUrl", data.uuid)
+      active = false
+    }
+
+    loading = false
+  }
+
+  async function renameProject() {
+    if (!$isSignedIn) {
+      alert("You must be signed in to rename a project")
+      active = false
+      return
+    } else if (!$currentProject) {
+      alert("No project selected? This is probably a bug.")
+      active = false
+      return
+    }
+
+    loading = true
+
+    const data = await renameCurrentProject(value)
+    if (data) active = false
+
+    loading = false
+  }
+</script>
+
+{#if active}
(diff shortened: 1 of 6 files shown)


app/javascript/src/utils/project.js
```diff
@@ -0,0 +1,127 @@
...
+    })
+    .catch(error => {
+      console.error(error)
+      alert("Something went wrong while creating your project, please try again")
+    })
+}
+
+export function createDemoProject() {
+  const newProject = {
+    uuid: Math.random().toString(16).substring(2, 8),
```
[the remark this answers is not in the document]
Mitcheljager: That's cool, but it poses no security risks here. Thanks though mr codescanning!

app/javascript/src/components/editor/Modals/CreateProjectModal.svelte
```diff
@@ -0,0 +1,94 @@
...
+    loading = false
+  }
+</script>
+
+{#if active}
+  <div class="modal modal--top" transition:fade={{ duration: 100 }} data-ignore>
+    <div class="modal__content p-0" transition:fly={{ y: 100, duration: 200 }}>
+      {#if !$isSignedIn}
+        <div class="warning warning--orange">
+          You are not signed in and this is for demonstration purposes only. This will not be saved.
```
CactusPuppy, line 57:
```suggestion
          You are not signed in and this is for demonstration purposes only. Any changes you make will not be saved.
```

app/javascript/src/components/editor/ProjectsDropdown.svelte
```diff
@@ -1,166 +1,63 @@
...
-      alert("You must be signed in to rename a project")
-      showModal = false
-      return
-    } else if (!$currentProject) {
-      alert("No project selected? This is probably a bug.")
-      showModal = false
-      return
-    }
+  async function duplicateProject() {
+    if (!confirm("This will create a copy of the current project. Please confirm!")) return
```
CactusPuppy, line 48:
```suggestion
    if (!confirm("This will create a copy of the current project. Do you want to continue?")) return
```

CactusPuppy (changes requested):
Most of my suggestions are for copy that was always there, but which I never got a chance to look at. The actually new features look good to me!

CactusPuppy (approved):
LGTM!

PR #289, opened 2023-04-14
Editor: @if directive

netux:
Syntax
```
@if (lefthand operator righthand) {
  put this
} @else {
  put this instead
}
```

where the `@else` is optional

- `lefthand` and `righthand` are (so far) only compared as raw strings
  - Currently supported operators are == and !=
- Runs after mixins, so it compares against Mixin values instead of `Mixin.XYZ`
- In classic `@mixin` fashion, there is no trimming of the content that gets replaced
  - but `lefthand` is right-trimmed and `righthand` is left-trimmed, for code convenience

Left a couple TODOs in the code, maybe we can tackle them in the code review.

Also, adjusted eslint rules so `switch` statements get an indent of 1 (the default in eslint is wrong and there is no way to make me change my mind about it)

### DIFF
--- a/app/javascript/src/utils/compiler.js
+++ b/app/javascript/src/utils/compiler.js
@@ -154,8 +154,8 @@ function extractAndInsertMixins(joinedItems) {
 }
 
 const conditionalOperations = {
-  "==": (l, r) => l === r,
-  "!=": (l, r) => l !== r
+  "is not": (l, r) => l !== r,
+  "is": (l, r) => l === r
 }
 
 function evaluateConditionals(joinedItems) {
--- a/app/javascript/src/utils/compiler.js
+++ b/app/javascript/src/utils/compiler.js
@@ -153,8 +153,13 @@ function extractAndInsertMixins(joinedItems) {
   return joinedItems
 }
 
+const conditionalOperations = {
+  "==": (l, r) => l === r,
+  "!=": (l, r) => l !== r
+}
+
 function evaluateConditionals(joinedItems) {
-  const conditionalStartRegex = /@if *\( *((?:.|\n)+?) *(==|!=) *((?:.|\n)+?)\) *[ \n]*\{/g
+  const conditionalStartRegex = new RegExp(`@if *\\( *((?:.|\\n)+?) *(${ Object.keys(conditionalOperations).join("|") }) *((?:.|\\n)+?)\\) *[ \\n]*\\{`, "g")
   const conditionalElseStartRegex = / *@else *\{/
 
   let match
@@ -185,20 +190,9 @@ function evaluateConditionals(joinedItems) {
       }
     }
 
-    let passed = null
     const sanitizedLeft = left.trimEnd()
     const sanitizedRight = right.trimStart()
-    switch (operation) {
-      case "==": {
-        passed = sanitizedLeft === sanitizedRight
-        break
-      }
-      case "!=": {
-        passed = sanitizedLeft !== sanitizedRight
-        break
-      }
-      // TODO: contains operator, regex operator?
-    }
+    const passed = conditionalOperations[operation]?.(sanitizedLeft, sanitizedRight)
 
     const finalContent = passed ? trueBlockContent : falseBlockContent
     joinedItems = replaceBetween(
--- a/app/javascript/src/utils/compiler.js
+++ b/app/javascript/src/utils/compiler.js
@@ -149,7 +149,6 @@ function evaluateConditionals(source) {
 
     const afterClosingBracketIndex = getClosingBracket(source, "{", "}", afterMatchedTextIndex - 2)
     if (afterClosingBracketIndex < 0) {
-      // TODO: warn user?
       continue
     }
 
@@ -158,7 +157,7 @@ function evaluateConditionals(source) {
     const trueBlockContent = source.substring(afterMatchedTextIndex, afterClosingBracketIndex)
     let falseBlockContent = ""
 
-    conditionalElseStartRegex.lastIndex = afterClosingBracketIndex - 1
+    conditionalElseStartRegex.lastIndex = afterClosingBracketIndex - 1 // set start position for the exec below
     const elseMatch = conditionalElseStartRegex.exec(source)
     if (elseMatch != null) {
       const afterElseMatchedTextIndex = elseMatch.index + elseMatch[0].length
@@ -166,7 +165,7 @@ function evaluateConditionals(source) {
       if (matchingClosingBracketForElseIndex > 0) {
         falseBlockContent = source.substring(afterElseMatchedTextIndex, matchingClosingBracketForElseIndex)
         conditionalEndingIndex = matchingClosingBracketForElseIndex
-      } // TODO: else warn user?
+      }
     }
 
     let passed = null
--- a/app/javascript/src/utils/compiler.js
+++ b/app/javascript/src/utils/compiler.js
@@ -153,9 +153,28 @@ function extractAndInsertMixins(joinedItems) {
   return joinedItems
 }
 
+const regexRegex = /^\/(.+)\/(\w*)$/
+
 const conditionalOperations = {
+  // order is deliberate as to not make RegExp work too much
   "is not": (l, r) => l !== r,
-  "is": (l, r) => l === r
+  "is": (l, r) => l === r,
+  "contains": (l, r) => l.includes(r),
+  "test": (l, r) => {
+    const match = r.match(regexRegex)
+    if (!match) {
+      // TODO: mark in linter
+      return false
+    }
+    const [_, pattern, flags] = match
+    let regex
+    try {
+      regex = new RegExp(pattern, flags)
+    } catch (err) {
+      return false
+    }
+    return regex.test(l)
+  }
 }
 
 function evaluateConditionals(joinedItems) {
--- a/app/javascript/src/utils/compiler.js
+++ b/app/javascript/src/utils/compiler.js
@@ -17,7 +17,7 @@ export function compile(overwriteContent = null) {
 
   joinedItems = joinedItems.replace(settings, "")
   joinedItems = extractAndInsertMixins(joinedItems)
-  joinedItems = extractAndInsertConditionals(joinedItems)
+  joinedItems = evaluateConditionals(joinedItems)
   joinedItems = convertTranslations(joinedItems)
 
   const variables = compileVariables(joinedItems)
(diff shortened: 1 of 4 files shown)


app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
...
+    switch (operation) {
+      case "==": {
+        passed = sanitizedLeft === sanitizedRight
+        break
+      }
+      case "!=": {
+        passed = sanitizedLeft !== sanitizedRight
+        break
+      }
+      // TODO: contains operator, regex operator?
```
netux, line 211: For these I was thinking
```
abc *= b     # compare using `String,includes`: `lefthand.includes(righthand)`
abc ~= /B/i  # parse regex and compare using `RegExp.test()`: `parsedRighthandRegex.test(lefthand)`
```

netux: In the case of the regex operator, what should happen if the righthand is not a valid regex? Make the operator silently return false? Warn to the user somehow?

app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
   return joinedItems
 }
 
+/**
+ * @param {string} source
+ * @param {number} fromIndex
+ * @returns
+ */
```
Mitcheljager, line 145: This is nice and all, but we don't do it anywhere else so let's not do it here either

app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
...
+    } else if (char === "}") {
+      currentDepth--
+      if (currentDepth <= 0) {
+        result = i
+        break
+      }
+    }
+  }
+  return result
+}
```
Mitcheljager, line 162: There already is a function that does the same `getClosingBracket` in `/utils/editor.js`.

app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
...
+
+/**
+ * @param {string} source
+ * @returns {string}
+ */
```
Mitcheljager, line 167: Same here, it's nice, but we don't do it anywhere else

app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
...
+function evaluateConditionals(source) {
+  const conditionalStartRegex = /@if *\( *((?:.|\n)+?) *(==|!=) *((?:.|\n)+?)\) *[ \n]*\{/g
+  const conditionalElseStartRegex = / *@else *\{/
```
Mitcheljager, line 170: One thing these 2 regexes miss is different formatting for the brackets. By default the Workshop puts brackets on the next line like
```
if (...)
{
  ...
}
else
{
  ...
}
```

I personally don't prefer that formatting at all, but we should support it. Perhaps something like `@else\s*\{`

app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
...
+  }
+  return result
+}
+
+/**
+ * @param {string} source
+ * @returns {string}
+ */
+function evaluateConditionals(source) {
+  const conditionalStartRegex = /@if *\( *((?:.|\n)+?) *(==|!=) *((?:.|\n)+?)\) *[ \n]*\{/g
```
Mitcheljager, line 169: Regex, man, what is even happening :D

app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
...
+  const conditionalElseStartRegex = / *@else *\{/
+
+  let match
+  while ((match = conditionalStartRegex.exec(source)) != null) {
+    const [matchedConditionalStartText, left, operation, right] = match
+    const afterMatchedTextIndex = match.index + matchedConditionalStartText.length
+
+    const matchingClosingBracketIndex = findMatchingClosingBracketIndex(source, afterMatchedTextIndex)
+    if (matchingClosingBracketIndex < 0) {
+      // TODO: warn user?
```
Mitcheljager, line 179: The warning would be in the linter, not here

app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
...
+
+    conditionalElseStartRegex.lastIndex = matchingClosingBracketIndex
+    const elseMatch = conditionalElseStartRegex.exec(source)
+    if (elseMatch != null) {
+      const afterElseMatchedTextIndex = elseMatch.index + elseMatch[0].length
+      const matchingClosingBracketForElseIndex = findMatchingClosingBracketIndex(source, afterElseMatchedTextIndex)
+      if (matchingClosingBracketForElseIndex > 0) {
+        falseBlockContent = source.substring(afterElseMatchedTextIndex, matchingClosingBracketForElseIndex)
+        conditionalEndingIndex = matchingClosingBracketForElseIndex
+      } // TODO: else warn user?
```
Mitcheljager, line 196: The warning would be in the linter, not here

app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
...
+    }
+
+    let passed = null
+    const sanitizedLeft = left.trimEnd()
+    const sanitizedRight = right.trimStart()
+    switch (operation) {
```
Mitcheljager, line 202: We don't use switch cases anywhere. Not really anything against it, but perhaps it would be better to not do it here either. Mostly because there's 2 cases here which might be much nicer as an if statement.
```
if (operation == "==") passed = sanitizedLeft == sanitizedRight
else if (opeartion == "!=") passed = sanitizedLeft != sanitizedRight
```

app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
...
+      case "==": {
+        passed = sanitizedLeft === sanitizedRight
```
Mitcheljager, line 204: Let's stick to `==` instead of `===` since we don't use it anywhere else. In this case it will only ever be strings any way.

app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
...
+        break
+      }
+      case "!=": {
+        passed = sanitizedLeft !== sanitizedRight
```
Mitcheljager, line 208: Let's stick to `!=` instead of `!==` since we don't use it anywhere else. In this case it will only ever be strings any way.

app/javascript/src/utils/compiler.js
Mitcheljager: Yep, I think in that case it should be silent. Any warnings should be in the linter.

app/javascript/src/utils/compiler.js
```diff
@@ -137,6 +138,93 @@ function extractAndInsertMixins(joinedItems) {
...
+    }
+  }
+  return result
+}
+
+/**
+ * @param {string} source
+ * @returns {string}
+ */
+function evaluateConditionals(source) {
```
Mitcheljager, line 168: This is a bit nitpicky, but instead of calling it `source` can we call it `joinedItems` to stay consistent with other functions?

Mitcheljager (changes requested):
Overall a great feature. It's might be a bit basic, but it's super useful anyway. In the future we might want to support `else if` and `&&` and `||`, and more, but for now this is a great start.

I left a lot of comments though, mostly about syntax.

---

I also have a proposal. What if we change the syntax slightly. Since we are only ever comparing strings, how about we don't use `==` and `!=`, but instead use words.

For example in SCSS a for loop looks like this: `@for $i from 1 through 4 {`, which I personally think looks very nice. What if our if statements looks like `@if (Mixin.thing is Mixin.otherThing) {` and `@if (Mixin.thing is not Mixin.otherThing) {`. This opens up the door for something like `@if (Mixin.thing contains Mixin.otherThing) {`
Perhaps the parenthesis could even be optional ``@if Mixin.thing is Mixin.otherThing {`, but other things (`@translate`) would have to follow in that case, so perhaps best to leave that for another time.

I don't think that would result in you having to change the code much either, right?

Let me know what you think!

app/javascript/src/utils/compiler.js
netux: https://regex101.com/r/WN1vYH/1

There is a lot of fluff for the spaces ` *`.
`((?:.|\n)+?)` matches the left hand or right hand.

app/javascript/src/utils/compiler.js
netux: How would one go about moving it to the linter?

app/javascript/src/utils/compiler.js
netux: I used these mainly for myself since my IDE wouldn't recognize types and show autocomplete otherwise. Do these work on your IDE?

I'd argue JSDocs are a necessity for non TypeScript projects. What if instead we start rolling these out gradually as we change things?

app/javascript/src/utils/compiler.js
netux: I used `switch` so in the future we could expand (see comment below the last `case`)

app/javascript/src/utils/compiler.js
netux: I just did `===` out of habit. I don't see why we strictly need to change this to `==`, unless you plan the lefthand and righthand to potentially be different types in the future?

app/javascript/src/utils/compiler.js
Mitcheljager: I personally haven't had an issue with it, but I can definitely see it being useful. In any case, I'd prefer to leave that for a different PR if we decide to do it at all

app/javascript/src/utils/compiler.js
netux: _Or_ we could use `source` on the other functions too 😁.
Honestly whichever way is fine for me 👍

app/javascript/src/utils/compiler.js
Mitcheljager: Perhaps a different solution could be something like
```js
const operations = {
  "==": (l, r) => l == r,
  "!=": (l, r) => l != r
 }
 
passed = operations[operation]?.(sanitizedLeft, sanitizedRight)
 ```
 
 Which could turn in to
 ```js
const operations = {
  "is": (l, r) => l == r,
  "is not": (l, r) => l != r,
  "contains": (l, r) => l.contains(r)
 }
 ```

app/javascript/src/utils/compiler.js
netux: Not doing this, as discussed on Discord: https://discord.com/channels/570672959799164958/574537911114924043/1096427037595537448 
image

PR #290, opened 2023-04-14
feat: Add duplicate for items and folders

Mitcheljager:
This PR add duplicating for items and folders. You can duplicate anything and an exact copy will be made right under the item. If it's a folder it will duplicate any items and folders that are inside of it as well. I've also added tooltips through data attributes to the buttons to explain what each button does.

duplicate-item-folder


app/assets/stylesheets/general/_general.scss
```diff
@@ -128,4 +128,24 @@ strong {
   font-size: $font-size-base;
   width: 100%;
   cursor: pointer;
+
+  &[data-tooltip]:hover {
+    position: relative;
+
+    &::after {
```
netux, line 135: Tooltip clips below Add button 
image

app/javascript/src/utils/editor.js
```diff
@@ -43,6 +43,19 @@ export function isAnyParentHidden(item) {
   return false
 }
 
+export function duplicateItem(item, newParent = null) {
+  const name = newParent ? item.name : item.name + " (Copy)"
```
netux, line 47: May be a little extra, but I think it's common for duplicated items to have a number increasing with each subsequent copy. Something like:
1. Check if there is an "{ item.name } (Copy)" under the same folder already
2. If so, use "{ item.name } (Copy 2)" instead
3. Check fi there is an "{ item.name } (Copy 2)" under the same folder already
4. If so, use "{ item.name } (Copy 3)" instead
5. Repeat as many times as necessary

Mitcheljager: Good idea, I've changed it.

duplicate-item-folder-count

PR #292, opened 2023-04-15
Editor: Parse conditional conditions manually

netux:
By itself this doesn't change anything, but it makes it very easy to expand what conditions can hold in the future


Mitcheljager (approved):
Very nice!

PR #293, opened 2023-04-15
feat: Add @for loop to editor compiler

Mitcheljager:
This PR adds `@for` to the compiler, allow you to create loops from within the editor. In a lot of cases for loops from within the Workshop will be preferred, but doing it in the compiler opens up some options that would not be possible using in-game for loops. For example you could use it to create a large number of rules with a couple of lines of code.

Some examples:

The most basic example, a loop running from 1 to 5, _including_ the 5. Hence it being `through` and not `to`. The iterator can be set to any string, in this case it's `someVar`, which can than be used like `For.someVar` to use it's value.
```scss
@for (someVar from 1 through 5) {
	For.someVar;
}
```

The iterator variable is optional. In this case the default value "i" will be used.
```scss
@for (0 through 10) {
	For.i;
}
```

Loops can be nested in other loops
```scss
@for (1 through 3) {
	@for (j from 2 through 5) {
		For.i + For.j
	}
}
```

`@if` can be used inside of loops
```scss
@for (0 through 10) {	
	@if (For.i is 5) {
		Im number 5: For.i
	}
	
	@if (For.i test /[6-9]|[1-9]\d+/) {
		Im higher than number 5: For.i
	}
}
```
Mixins can be inside of loops too, in this case both Mixin.param and For.i will be the same
```scss
@for (5 through 6) {
	@include someRule(For.i) {
		For.i;
		Mixin.param;
	}
}
```
They can also be used inside of Mixins and take a mixin param as one of the values
```scss
 @mixin someMixin(someInt = 5) {
	@for (someVar from 1 through Mixin.someInt) {
		For.someVar;
    }
}
```


netux:
How about `to` instead of `through`? Easier to type.

I'm also thinking `@loop` instead of `@for` sounds more natural. "Loop (as) someVar from 1 to 9" vs. "For (each) someVar from 1 to 9". On the other side `@for` is more recognizable for, well, __for__ loops.

Any plans to reuse `@for` for other types of loops (mostly probably a for each)?

netux (approved):
Simple but effective. LGTM!

Mitcheljager:
@netux I went with `through` because that implies the number is inclusive, where as `to` would be exclusive. However, I've now added both.
```scss
@for 1 through 3 { For.i } // 1 2 3
@for 1 to 3 { For.i } // 1 2

---

As for naming it `@loop`, I thought about that too, it is a nicer name. But I decided against it because it would be different from both SCSS and the Workshop itself, where for loops are just like this.
I don't think `@for` would be re-used for other loops. `@each` would be used for each loops I think.

File: app/javascript/src/utils/compiler.js [hunk 1]
```diff
@@ -324,3 +325,29 @@ function convertTranslations(joinedItems) {
 
   return joinedItems + rule
 }
+
+function evaluateForLoops(joinedItems) {
+  let match
+  const forRegex = /@for\s+\(\s*((?:(\w+)\s+)?(?:from\s+))?(\d+)\s+(?:(through|to)\s+)?(\d+)\s*\)\s*\{/g // Matches "@for ([var] [from] number through|to number) {" in groups for each param
+  while ((match = forRegex.exec(joinedItems)) != null) {
+    console.log(match)
```
netux, line 333: Remove comment

```suggestion
```

app/javascript/src/utils/compiler.js
```diff
@@ -331,9 +331,9 @@ function evaluateForLoops(joinedItems) {
   const forRegex = /@for\s+\(\s*((?:(\w+)\s+)?(?:from\s+))?(\d+)\s+(?:(through|to)\s+)?(\d+)\s*\)\s*\{/g // Matches "@for [var] [from] number through number" in groups for each param
   while ((match = forRegex.exec(joinedItems)) != null) {
     console.log(match)
-    const [full, _, variable, start, to, end] = match
+    const [full, _, variable, start, clusivity, end] = match
```
netux, line 334: I think you meant `inclusivity`.

Personally I would have called it `toAdverb`

Mitcheljager: Clusivity is a real word
> Back-formation from inclusivity and exclusivity.

netux (approved):

PR #295, opened 2023-04-16
Configure VSCode to fix all ESLint problems on .svelte and .js file save

netux:


Mitcheljager:
I don't think I really like this to be honest. It's of course nice to have things fixed before you push them, but I would personally find this rather annoying.

netux:
This was the easiest way to fix auto-imports not respecting ESLint rules.
The other option was configuring settings for semi, quotes, etc. manually to match the ESLint ones.

Auto linting sounds annoying at first, until you realize it actually saves you from having to notice these errors in the first place.
Auto formatting with prettier is the frustrating one imo. No one is ever completely comfortable with the style it imposes.

Mitcheljager:
I think if you prefer it it's better to add to your own VSCode config, rather than enforcing it on everyone.

PR #296, opened 2023-04-16
Editor: Move Add button to be an icon next to the "Items" title

netux:
image

Also refactors some code to make creating buttons with icons easier.


app/assets/stylesheets/elements/_buttons.scss
```diff
@@ -137,4 +137,25 @@
   .spinner {
     margin-bottom: $margin / 8;
   }
+
+  &--icon-square {
+    @extend .button--square;
```
Mitcheljager, line 142: Extend should only be used when applying the class itself doesn't make sense or is not easily possible. For example if you have an element `.something__child`, if you want to reuse that in a block `.else__child`, you wouldn't be able to add `something__child`, because that would be a different block. You would also only use `@extend` if it the styling has to be the exact same in all cases. `@extend` create ugly selector stacks that would be good to avoid where possible.

In this case you should apply the class `button--square` separately. But in that case the name `icon-square` doesn't make sense, since it's the  class `buton--square` that does the squaring. So just `icon` might suffice.

app/javascript/src/components/editor/EditorItemCreate.svelte
```diff
@@ -32,13 +33,13 @@
 
 <svelte:window on:click={outsideClick} on:keydown={event => { if (event.key === "Escape") active = false }} />
 
-<div class="dropdown w-100" class:dropup>
-  <button class="empty-button" on:click|stopPropagation={toggle} bind:this={element}>
-    <slot />
```
Mitcheljager, line 37: This was using a slot so it could in the future be re-used in other places, for example create an item directly in a folder.

app/javascript/src/components/editor/Settings.svelte
```diff
@@ -72,7 +72,7 @@
 <svelte:window on:click={outsideClick} on:keydown={event => { if (event.key === "Escape") active = false }} />
 
 <div class="dropdown settings">
-  <button class="button button--secondary button--square" on:click|stopPropagation={() => active = !active}>
+  <button class="button button--secondary button--icon-square" on:click|stopPropagation={() => active = !active}>
```
Mitcheljager, line 75: See naming comments above

app/javascript/src/components/editor/Shortcuts.svelte
```diff
@@ -39,7 +39,7 @@
 <svelte:window on:click={outsideClick} on:keydown={event => { if (event.key === "Escape") active = false }} />
 
 <div class="dropdown shortcuts settings">
-  <button class="button button--secondary button--square" on:click|stopPropagation={() => active = !active}>
+  <button class="button button--secondary button--icon-square" on:click|stopPropagation={() => active = !active}>
```
Mitcheljager, line 42: See naming comments above

app/javascript/src/components/icon/Plus.svelte
```diff
@@ -0,0 +1,3 @@
+<svg viewBox="0 0 32 32" width="32px" height="32px" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg" class="fill-current-color">
```
Mitcheljager, line 1: Since the SVG is inline most of the fluff can be remove. `version`, `xmlns`, `sxmlns:svg` are are not necessary.

Mitcheljager (changes requested):
I left a couple comments, mostly concerning `@extend`. But other than that I also have some usability concerns. The button is at the bottom because that is where items go when they are created. With the button being at the top you will not see the item being created if it is out of view. Overall I'm not sure what this PR is trying to achieve or improve.

Mitcheljager:
One more addition; the copy for empty projects would need to be changed, as it currently directs you to "the button below"

netux:
My goal was to match other editors (namely VSCode), which have a toolbar at the top of the file tree with buttons to create files, expand all, etc. Also making it easier to reach as you wouldn't have to scroll down when the file list is too large.

Your point about having visual feedback on where the file is after being created is interesting. I would have suggested scrolling the file into view, but I also just noticed the file list doesn't have a separate scrollbar from the rest of the sidebar, so the "Items" header and now the create file button would go out of view 🤔.

So yeah, you are right, this solves nothing currently. We can discuss it better on Discord.

PR #297, opened 2023-04-16
Editor: Fix crash when evaluating an empty expression

netux:
Fixes the `TypeError: Cannot read properties of undefined (reading 'map')` when there is an `@if` with an empty condition like so:
```
@if () {}
```

This has to be handled separately because, when parsing the expression, since there are no characters the for loop would just go to the end, returning an empty object.


Mitcheljager (approved):
Nice!

PR #298, opened 2023-04-16
Editor: Add lint rule for @else if uses

netux:
Not sure when or if we will ever implement `@else if`, but this should prevent some confusion.


Mitcheljager (approved):
Nice and simple!

PR #299, opened 2023-04-18
Editor: Fix some issues with the demo project when not signed-in

netux:
- Fix crash creating demo project in the first place
- Fix Save and Import buttons showing for non signed-in users
- Hide Edit dropdown for non signed-in users, as they cannot use any of the buttons in there


Mitcheljager (approved):
Ah whoops, good catch!

Mitcheljager (approved):

PR #300, opened 2023-04-18
Add VSCode launch task for Rails server

netux:
Allows VSCode users to set breakpoints and inspect variables.

Turns out this was pretty easy: I just needed to use the template configuration, and to install gems and an extension.


Mitcheljager (approved):
Sure 👍

PR #302, opened 2023-04-19
Editor: @each loop

netux:
Another late night, another workshop.codes/editor directive

**Syntax:**
```
@each (value in [a, b, c]) {
  Each.value (#Each.i)
}
// results in
  a (#0)
  b (#1)
  c (#2)

@each (value, index in [a, b, c]]) {
  Each.value (#Each.i)
}
// results in the same as above

@each (value in Button) { // or any other type of Workshop Constant from the wiki
  Each.value
}
// results in
  Ability 1
  Ability 2
  ...

@each (value in Buttons) { // "plurals", but done in a very naïve and English-only way
  Each.value
}
// results in the same as above
```

**Some notes:**
- Workshop Constants only work in English. Ideally we'd have some setting to select the language, but that's out of scope for now. Also, the wiki resource this uses doesn't have non-english strings for the names of the constants.
- Arrays are parsed in a way so things inside parentheses are evaluated as single items. E.g.: parsing "[1, (2, 3), 4]" would be the JavaScript array `["1", "(2, 3)", "4"]`, and not `["1", "(2", "3)", "4"]`
  - Because this is not a simple `iteratableStr.split(/, +/)`, and due to the implementation, there are some edge cases in which we accept empty items (e.g. "[a,,c]" would be parsed as the JS array `["a", "", "c"]`) except when they are the last item of the array (e.g. "[a,b,]" would be parsed as `["a", "b"]`, without the expected empty string at the end). I didn't want to bother much with it yet, but it is a thing to improve in the future for sure.
- No syntax highlight or special linter support yet. I don't want to leave this PR open for so long like last time, so we'll add that later.


app/javascript/src/utils/compiler.js
```diff
@@ -410,3 +411,101 @@ function evaluateForLoops(joinedItems) {
...
+    let iterable = []
+    if (iterableStr[0] === "[" && iterableStr[iterableStr.length - 1] === "]") {
+      iterable = parseArrayValues(iterableStr.substring(1, iterableStr.length - 1))
+    } else {
+      const language = get(defaultLanguage)
+      const constants = get(workshopConstants)
+
+      const sanitizedIterableStr = iterableStr.toLowerCase().trim()
+
+      // TODO: support constant names in other languages
```
Mitcheljager, line 464: Other languages won't be supported, they aren't supported anywhere else either.

app/javascript/src/utils/compiler.js
```diff
@@ -410,3 +411,101 @@ function evaluateForLoops(joinedItems) {
...
+        })
+
+      if (usedConstantName != null) {
+        iterable = Object.values(constants[usedConstantName]).map((value) => value[language])
+      }
+    }
+
+    if (iterable == null) {
+      continue
+    }
```
Mitcheljager, line 480: This can be on one line

```suggestion
if (iterable == null) continue
```

app/javascript/src/utils/compiler.js
```diff
@@ -410,3 +411,101 @@ function evaluateForLoops(joinedItems) {
...
+    const indexVarRegex = new RegExp(`Each.${ indexVar || "i" }(?=\\W|$)`, "g")
+    const valueVarRegex = new RegExp(`Each.${ valueVar }(?=\\W|$)`, "g")
+
+    let finalContent = ""
+    // Replace "Each.[valueVar]" and "Each.[indexVar]"
+    for (const [index, value] of Object.entries(iterable)) {
+      finalContent += contentToRepeat
+        .replaceAll(indexVarRegex, index)
+        .replaceAll(valueVarRegex, value)
+    }
```
Mitcheljager, line 498: To be a little more consistent it might be better to use .forEach or .reduce, `for...of` isn't used anywhere I believe.

Something like this might be nice enough
```suggestion
const finalContent = Object.entries(iterable).reduce((current, [index, value]) => {
  return current + contentToRepeat
    .replaceAll(indexVarRegex, index)
    .replaceAll(valueVarRegex, value)
}, "")
```
Or otherwise a regular for loop.

app/javascript/src/utils/compiler.js
```diff
@@ -410,3 +411,101 @@ function evaluateForLoops(joinedItems) {
...
+      nextStartingIndex = lastValidCommaEndIndex + 1
+    }
+
+    commaRegex.lastIndex = nextStartingIndex
+  }
+
+  const lastValue = input.substring(lastValidCommaEndIndex + 1)
+  if (lastValue.length > 0) {
+    result.push(lastValue)
+  }
```
Mitcheljager, line 444: Let's one-line this

```
if (lastValue.length > 0) result.push(lastValue)
```

Mitcheljager (changes requested):
Very cool! I love the addition of constants. I left some small nitpicks all related to formatting, but great job overall!

app/javascript/src/utils/compiler.js
netux: Fair enough. But what if they were? 🤔

Ok ok, I will remove the TODO.

Mitcheljager: Maybe one day when someone actually asks for it. But for now the amount of maintenance required far outweighs the benefits

PR #303, opened 2023-04-19
style: Post show design changes

Mitcheljager:
This PR adjusts the design of show pages in fairly subtle ways. Each section is divided in blocks, with the intend to make it a little more well organized, a little more open, and it looks nice.

## Overall
| Before | After |
| --- | --- |
| image | image |

## Banners
| Before | After |
| --- | --- |
| image | image |

## Without carousel
| Before | After |
| --- | --- |
| image | image |

## No description
| Before | After |
| --- | --- |
| image | image |

## Derivs, expire soon notice
| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/app/assets/stylesheets/_variables.scss
+++ b/app/assets/stylesheets/_variables.scss
@@ -8,6 +8,7 @@ $orange: #e89930;
 $green: #3fbf74;
 $body-bg: #1f2326;
 $primary: var(--primary);
+$primary-var: #f06414; // FIXME: This sucks but is used for scss methods. Should refactor $primary to not be a css variable
 $bg-dark: #2e3337;
 $bg-darker: var(--bg-darker);
 $text-color: var(--text-color);
@@ -56,7 +57,7 @@ $media-min-widths: (
 
 :root {
   color-scheme: dark;
-  --primary: #f06414;
+  --primary: #{ $primary-var };
   --white: #dfdfdf;
   --orange: #{ $orange };
   --red: #{ $red };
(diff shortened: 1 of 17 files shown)


CactusPuppy:
I think the tab selector block should still be attached to the block below it to make it clear what element the tab selector controls.

Mitcheljager:
@CactusPuppy I see what you mean, but I don't know how I could do this in the current style and still have it look good. I think it's clear enough as is either way.

Mitcheljager:
Inspired by the YouTube video glow thing I've added a glow effect around the carousel based on the average color of the current image.

_The gif does not do it justice, it looks very artifacty here, but that's not actually the case_

carousel-glow

CactusPuppy:
> @CactusPuppy I see what you mean, but I don't know how I could do this in the current style and still have it look good. I think it's clear enough as is either way.

Agreed, seeing it in action makes it look less awkward than it looked at first glance.

Mitcheljager:
@CactusPuppy I ended up changing it a little bit with what you said in mind;

The tabs are now inside of the title block, which is more in line with other pages (namely the profile pages).

image

CactusPuppy:
> @CactusPuppy I ended up changing it a little bit with what you said in mind;
> 
> The tabs are now inside of the title block, which is more in line with other pages (namely the profile pages).
> 
> image

I like this as well. I think my main gripe looking back was the isolation of the tab selection block, which this solves @Mitcheljager

Mitcheljager:
@CactusPuppy I agree, it was a bit much

PR #304, opened 2023-04-20
Editor: @each v2, improvements and changes

netux:
- Add linter for invalid iterables
- Add highlight for terms used inside `@each`
- Change Workshop Constant support to require an explicit "Constant." prefix, and no longer allow plurals
- Fix edge-case with array content parsing
- Take nested arrays inside `@each` array iterables as whole items, allowing replacing them in nested `@each`s. Example in comment of 0e2dad6.


app/javascript/src/lib/OWLanguageLinter.js
```diff
@@ -509,3 +510,43 @@ function findConditionalsElseIfUses(content) {
...
+    if (constantMatch != null) {
+      const [constantFull, constantName] = constantMatch
+
+      if (!constants[constantName]) {
+        const from = match.index + constantMatch.index
+        diagnostics.push({
+          from,
+          to: from + constantFull.length,
+          severity: "error",
+          message: `\"${ constantName }\" is not a known Workshop Constant`
```
Mitcheljager, line 538: You don't need to escape the quotes here since it's in a template literal

Mitcheljager (approved):
One small comment, but overall this is great!

app/javascript/src/lib/OWLanguageLinter.js
netux: Good point

```suggestion
          message: `"${ constantName }" is not a known Workshop Constant`
```

app/javascript/src/utils/compiler.js
```diff
@@ -412,6 +412,17 @@ function evaluateForLoops(joinedItems) {
...
+  "(": ")",
+  "[": "]"
+}
+const openBracketRegex = new RegExp(
+  Object.keys(openToClosingBracketsMap)
+    .map((c) => `\\${ c }`)
+    .join("|"),
+  "g"
+)
+
```
Mitcheljager, line 425: Let's put consts like this either in the function they belong in or in a separate file, otherwise they might get lost

netux: I didn't want to put them inside the function since it could be called more than once per `@each` match, and I didn't think they'd be good in a separate file since they are specific to `parseArrayValues()` (e.g. there is no `"{": "}"` entry)

Mitcheljager: I see, let's put them at the top of the file then

netux: How about having this file structure:
```js
./compiler
├── array.js
│   ├── const openToClosingBracketsMap
│   ├── const unescapedOpenBracketRegex
│   ├── const escapedCharRegex
│   └── parseArrayValues(input)
├── expression.js
│   ├── const comparisonOperators
│   ├── const sortedComparisonOperatorsSymbols
│   ├── getExpressionTree(expression)
│   └── evaluateExpressionTree(tree)
└── index.js // before: compiler.js
```

only issue is where to put shared functions like `removeSurroundingParenthesis(source)` (a shared.js?)

Mitcheljager: How about separating it even more?
- Compiler
  - Mixin.js
  - For.js
  - Each.js
  .....
  - helpers.js

Mitcheljager: Let's discuss it later some more, not in this PR

PR #306, opened 2023-04-22
feat: Improve data provided by report notifications

Mitcheljager:
With this PR notifications send to Discord will provide more data with actionable links right from the embed. The notify_discord action is split up in to two different ones, one for posts and one for comments. They could probably have been combined in some smart way, but this felt like the cleaner (and easier) solution.

Posts contain info about the post itself. Namely the title, code, image, if it has a snippet, and the start of a description. All of which are useful when determining if a code is real.

Comments contains the comment body itself. There are no links set up to view the comment in the admin or on the post itself. We rarely (if ever) get comment reports so I decided not to spend time on it for now.

Both have Accept and Reject buttons, which are really just links, but they do accept or reject the report as you'd expect.

| Posts | Comments |
| --- | --- |
| image | image |

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -106,6 +106,7 @@ Rails.application.routes.draw do
     delete "linked_users/:id", to: "linked_users#destroy", as: "destroy_linked_user"
 
     resources :reports, only: [:create, :new, :destroy, :update]
+    get "reports/:id/:status", to: "reports#update", as: "report_submit"
     resources :notifications, only: [:index], concerns: :paginatable
     get "unread-notifications", to: "notifications#get_unread_notifications"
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -106,6 +106,7 @@ def self.matches?(request)
     delete "linked_users/:id", to: "linked_users#destroy", as: "destroy_linked_user"
 
     resources :reports, only: [:create, :new, :destroy, :update]
+    get "reports/:id/:status", to: "reports#update", as: "report_submit"
```
CactusPuppy, line 109: This creates the possibility for a CSRF attack, but the potential consequences are minor.

CactusPuppy (approved):
LGTM

PR #313, opened 2023-06-17
Make links to headings/sections of an wiki article work in the Editor wiki sidebar

netux:
- Render header anchors on the server side
- Manually handle jumping to headings

### DIFF
--- a/app/javascript/src/components/editor/EditorWiki.svelte
+++ b/app/javascript/src/components/editor/EditorWiki.svelte
@@ -13,17 +13,13 @@
     const href = new URL(event.target.href)
 
     const currentHost = window.location.hostname
-    if (href.hostname === currentHost) {
-      if (!!href.hash) {
-        jumpToSection(href.hash.substring(1))
-        return
-      } else if (href.pathname.startsWith("/wiki")) {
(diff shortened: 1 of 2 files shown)


app/javascript/src/components/editor/EditorWiki.svelte
```diff
@@ -3,15 +3,31 @@
...
+      if (!!href.hash) {
+        jumpToSection(href.hash.substring(1))
+        return
+      } else if (href.pathname.startsWith("/wiki")) {
+        fetchArticle(event.target.href)
+        return
+      }
+    }
+
+    window.open(event.target.href, "_blank")
```
Mitcheljager, line 26: If we invert this check we could use an early return and indent less

```suggestion
    if (href.hostname != currentHost) {
      window.open(event.target.href, "_blank")
      return
    }
    
    if (!!href.hash) jumpToSection(href.hash.substring(1))
    else if (href.pathname.startsWith("/wiki")) fetchArticle(event.target.href)
```

Mitcheljager (changes requested):
Nice solution, good thinking! One small suggestion to make it a little more readable, but it's nice either way

PR #316, opened 2023-06-18
Fix some styling issues with the Editor line finder expanded snippet

netux:
- Use correct CSS classes for the snippets
- Move scrollbar to the expanded snippet, not the whole page
  - Not to happy about how I did this. I had to work around the fact modal is a flex parent. Open to suggestions.
- Put highlighted line into view expanding snippet

### DIFF
--- a/app/javascript/src/components/editor/ExpandableSnippet.svelte
+++ b/app/javascript/src/components/editor/ExpandableSnippet.svelte
@@ -10,12 +10,14 @@
   let expanded = false
   let snippetLines
   let snippetStartLineIdx
-  let expandedCodeEl
+  let expandedCodeElement
 
   $: {
     snippetStartLineIdx = snippetHighlightedLineIndex - Math.max(0, Math.floor(snippetLineCount / 2))
     snippetLines = getLines(snippetLineCount)
   }
+  $: syntaxHighlight(expanded)
+  $: if (expanded) scrollToHighlightedLine()
 
   function getLines() {
     const lines = []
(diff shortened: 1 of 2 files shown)


app/javascript/src/components/editor/ExpandableSnippet.svelte
```diff
@@ -10,6 +10,7 @@
   let expanded = false
   let snippetLines
   let snippetStartLineIdx
+  let expandedCodeEl
```
Mitcheljager, line 13: I'd prefer if it's not abbreviated. `expandedCodeElement` reads nicer

app/javascript/src/components/editor/ExpandableSnippet.svelte
```diff
@@ -33,13 +34,26 @@
...
+
+    if (!expandedCodeEl) return
+
+    const highlightedLineEl = expandedCodeEl.querySelector(".expandable-snippet__line--highlighted")
+    if (!highlightedLineEl) return
+
+    highlightedLineEl.scrollIntoViewIfNeeded()
+  }
+
+  $: if (expanded) scrollToHighlightedLine()
```
Mitcheljager, line 48: Would be nice if this was moved up to be with the other reactive functions. It doesn't have to be declared after the function, it gets hoisted anyway.

Mitcheljager (changes requested):
2 Small suggestions, but good otherwise!

app/javascript/src/components/editor/ExpandableSnippet.svelte
netux: It was more about keeping related things close to each other for readability, but sure 👍

PR #320, opened 2023-06-23
Fix a couple of a cases of splitting parameters not being aware of strings or parenthesis

netux:
Use `splitArgumentsString` in places where it should or could be used. On this PR:
- when splitting arguments in the linter
- when splitting parameters of a mixin


Mitcheljager (approved):
Sometimes it's that easy! Nice one

netux:
All thanks to this `splitArgumentsString` util generously provided by a very nice dev 😄

PR #321, opened 2023-06-23
Use new operator symbol for regex conditional check in Editor linter

netux:
`test` → `~=`

As if anyone was using the test operator anyway lol.


Mitcheljager (approved):
Sure, more consistent

PR #330, opened 2023-07-26
Add Editor linter to check for event types with missing arguments

netux:
Checks for `event {}` blocks with an event that requires arguments (e.g. Subroutine, Ongoing - Each Player), and warns the user if some arguments are missing.

In the future, we may want to get the list of arguments from `config/arrays/wiki/events.yml`.


Mitcheljager (approved):
Great addition!

PR #332, opened 2023-07-27
Prevent Editor linter crash when event type is not in the list of known events with arguments

netux:
This little bug disabled linting for the entire file. Oops.

### DIFF
--- a/app/javascript/src/lib/OWLanguageLinter.js
+++ b/app/javascript/src/lib/OWLanguageLinter.js
@@ -569,7 +569,7 @@ function findEventBlocksWithMissingArguments(content) {
 
     const requiredEventArgs = eventTypeToArgumentsMap[eventType.toLowerCase()]
 
-    if (requiredEventArgs.length !== givenEventArgs.length) {
+    if (requiredEventArgs && requiredEventArgs.length !== givenEventArgs.length) {
       const missingArguments = requiredEventArgs.slice(givenEventArgs.length)
       diagnostics.push({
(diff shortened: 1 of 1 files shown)


app/javascript/src/lib/OWLanguageLinter.js
```diff
@@ -569,7 +569,7 @@ function findEventBlocksWithMissingArguments(content) {
 
     const requiredEventArgs = eventTypeToArgumentsMap[eventType.toLowerCase()]
 
-    if (requiredEventArgs.length !== givenEventArgs.length) {
+    if (requiredEventArgs && requiredEventArgs.length !== givenEventArgs.length) {
```
Mitcheljager, line 572: Let's use optional chaining
```suggestion
    if (requiredEventArgs?.length !== givenEventArgs.length) {
```

Mitcheljager (changes requested):

app/javascript/src/lib/OWLanguageLinter.js
netux: ❌ Bad.
If `requiredEventArgs == null` and `givenEventArgs = [...]`, then `requiredEventArgs?.length = null`, then `requiredEventArgs?.length !== givenEventArgs.length => null !== [...] => true`.

Optional chaining here means we always go into the if block when we don't have a record of the event type.

Mitcheljager: Sorry I have no idea what you mean

PR #333, opened 2023-07-27
Fix some issues with the parsing of @each in the Editor

netux:
1. Allow array literals to have new lines

    ```haskell
    @each (value in [
      one,
      two,
      three
    ]) {
      Each.pair
    }
    ```
1. Fix detection regex overconsuming when there are multiple loops with array literals in the file

    showcase

1. Add a test for nested loops


Mitcheljager (approved):
Nice! Love that you included tests

PR #334, opened 2023-07-28
Fix the Editor script importer so it is able to find rules with spaces before the name

netux:
This fixes an issue when importing Workshop code generated via OverPy.

I've also simplified the regex for splitting rules a bit.


Mitcheljager (approved):
Nice!

PR #335, opened 2023-07-28
Replace keyCode → key throughout the Editor code

netux:
`KeyboardEvent.key` has been a standard for years. There is no reason to use `.keyCode` anymore.

This improves readability and allows for searching for keybinds more easily.

This PR also enhances the use === instead of == (because why wouldn't you use triple equals unless in very specific cases).


Mitcheljager (approved):
I don't really care which one we use, but .key sure is more readable!

PR #336, opened 2023-07-28
Improvements to aid IntelliSense (type hints, autocomplete, etc.) in VSCode

netux:
- Add popular Jest extension to the recommended list
- Install @types/jest as a dev dependency
- Add jsconfig.json with target lib ES2022 to silence some warnings


Mitcheljager (approved):
Sure!

PR #337, opened 2023-07-28
Add Ctrl+Shift+Z as a shortcut for redoing the last action in the Editor

netux:
Some programs use Ctrl+Y, others use Ctrl+Shift+Z. I say: why not both?


Mitcheljager (approved):
Sure, why not!

PR #338, opened 2023-07-28
Add Editor linter to check for usages of subroutines that aren't defined anywhere on the project

netux:
showcase


Mitcheljager:
Unfortunately this is not something we can do. The linter needs to be very quick, and looping over the entire project content is just not an option

netux:
@Mitcheljager I have changed it to retrieve the subroutines from the store, as they were already being extracted by the parser for autocomplete. I think then this should have a minimal impact on performance.

We could even try debouncing the subroutine extraction like you did in #339 for extra gains.

Mitcheljager:
@netux Nice, I like that. Let's merge this as is, and we will see if there's any obvious performance issues

PR #339, opened 2023-07-28
Rework variable extraction in the Editor, debounce autocomplete suggestion extraction

netux:
- Ignore variables inside strings
- Detect variables used in Set, Modify, and Chase Variable actions
- Overhaul detection of player variables
  - Any variable delimited with a period counts as a player variable
  - Detect nested player variables. E.g.:
    ```
    Event Player.pVar1.pVar2;
    Global.gVar1.pVar1; // always assumes gVar1 is a Player
    ```
- Update tests with new features and edge cases

Aside from the unit tests, I have also tested this by importing and then recompiling the following real-world game modes: [JHEM2 v1.3.0](https://workshop.codes/JHEM2), [1DMTZ v4.0.8](https://workshop.codes/1DMTZ), [7S44N v1.1.1](https://workshop.codes/7S44N), and [PZDS2 v1.0.3](https://workshop.codes/PZDS2). All variables matched, except for variables that were given names but went unused.

### DIFF
--- a/spec/javascript/utils/compiler/variables.test.js
+++ b/spec/javascript/utils/compiler/variables.test.js
@@ -8,14 +8,16 @@ describe("variables.js", () => {
         Global.variable1 = Test;
         Set Global Variable(variable2, Test);
         Modify Global Variable(variable3, Test);
-        Chase Global Variable At Rate(variable4, Test, ...);
-        Chase Global Variable Over Time(variable5, Test, ...);
-        For Global Variable(variable6, 0, 1) {
+        Set Global Variable(variable4, Test);
+        Modify Global Variable(variable5, Test);
+        Chase Global Variable At Rate(variable6, Test, ...);
+        Chase Global Variable Over Time(variable7, Test, ...);
+        For Global Variable(variable8, 0, 1) {
           // Do something
         }
       `
       const expectedOutput = {
-        globalVariables: ["variable1", "variable2", "variable3", "variable4", "variable5", "variable6"],
+        globalVariables: ["variable1", "variable2", "variable3", "variable4", "variable5", "variable6", "variable7", "variable8"],
         playerVariables: []
       }
       expect(getVariables(input)).toEqual(expectedOutput)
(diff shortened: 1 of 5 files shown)


spec/javascript/utils/compiler/variables.test.js
```diff
@@ -47,6 +79,49 @@ describe("variables.js", () => {
       expect(getVariables(input)).toEqual(expectedOutput)
     })
 
+    describe("Player variable edge cases", () => {
```
Mitcheljager, line 82: We can do without this `describe` in my opinion, currently they are used to denote files and function names

Mitcheljager (changes requested):
This is cool, but I have one fear. Currently the variables are extracted in a derived store to add them to the completions map. This is done when the content of the editor changes, which of course is on every keystroke. For this to run on every keystroke seems a bit insane for larger projects. We might need to debounce it to make sure it only runs when you stop typing. Maybe the debounce could be directly in the derived store itself, but I never tried.

What do you think? Have you done any tests?

netux:
> This is cool, but I have one fear. Currently the variables are extracted in a derived store to add them to the completions map. This is done when the content of the editor changes, which of course is on every keystroke. For this to run on every keystroke seems a bit insane for larger projects. We might need to debounce it to make sure it only runs when you stop typing. Maybe the debounce could be directly in the derived store itself, but I never tried.

Seems a little out of scope for this PR (and I wouldn't know how to apply a debounce on a derived Svelte store), so maybe we could tackle that later.

> What do you think? Have you done any tests?

No. All my tests were only on the logic, not the performance.

spec/javascript/utils/compiler/variables.test.js
netux: Would adding a prefix like "Edge-case: " be fine?

Mitcheljager: I don't think it's needed, tests are supposed to catch edge cases by definition. I like the idea, but I'd rather keep the definitions more consistent.

Mitcheljager:
@netux I will test it myself and make sure it doesn't make any obvious adverse effects. Debouncing it should be pretty easy, so if it's needed I can add a commit

netux:
I forgot to put the Co-Author, but should be done. Thanks!

netux:
I have merged the master branch with #338 in it, which splits `variablesMap$` into `variablesMap$` and `subroutinesMap$`.
So I went ahead and debounced all three `variablesMap$`, `subroutinesMap$`, and `mixinsMap$`.

app/javascript/src/utils/store.js
```diff
@@ -0,0 +1,8 @@
+export function debounced(cb, ms) {
+  return (values, set) => {
+    const timeout = setTimeout(() => {
+      return set(cb(values))
+    }, ms)
+    return () => clearTimeout(timeout)
+  }
+}
```
Mitcheljager, line 8: There actually already is a `debounce` util in `app\javascript\src\debounce.js`. Not sure if that would be suitable to re-use here.

netux: I tried it like so
```suggestion
import debounce from "../debounce"

export function debounced(cb, ms) {
  return debounce((values, set) => {
    return set(cb(values))
  }, ms)
}
```

but I was getting type errors because `get(subroutinesMap)` in the linter was `undefined`. I'm guessing there is some race condition somewhere but I don't know where to look.

It's probably better to just use a custom one tailored for `derived`.

Mitcheljager: Oh actually nevermind, that isn't the same of course. The set and return here are very important

app/javascript/src/utils/store.js
```diff
@@ -0,0 +1,8 @@
+export function debounced(cb, ms) {
```
Mitcheljager, line 1: Let's rename `cb` to something else, I have no idea what it stands for

Mitcheljager: Additionaly let's rename the file, it being under `utils/store` is a bit confusing to me. It can be it's own file, and with a more defined name `debouncedStore`, so it's clear this is only to be used for stores.

netux: `cb` stands for "callback", `ms` stands for `milliseconds`

Sure! How about
```suggestion
export function debounced(func, wait) {
```
to match the ones in `src/debounce.js`?

netux: As for the file name... maybe `storeUtils.js`? But it is already under `utils/`, so that is a bit redundant.

Mitcheljager: Ah callback, that's a good name actually. Let's call it that but not abbreviated.

I don't mind if the file is singlepurpose and has a name to match. It doesn't need to contain more store utilities

PR #340, opened 2023-07-29
feat: Adjust editor empty state

Mitcheljager:
This PR adjusts the initial empty state of the Editor, which appears when you are logged out, don't have any projects, or by pressing the logo in the top left.
There are multiple purposes behind these changes;
- The empty state still stated the editor was in Alpha. While this isn't false, it's also a bit meaningless at this point.
- Several of the known bugs were outdated
- It wasn't very actionable

Additionally the wiki sidebar is now always visible.

## Logged out
Showing 2 clear actions;
- "Create your first project" will create a demo project.
- "Check out sample project" will take you to my "Fishing in Antarctica", which is a nice enough demonstration of the editor.
image

## Logged in without any projects
This time the "Create your first project" button will create an empty project with the title "My First Project". One click, no need to input anything yourself. Otherwise this view is identical to the logged out view
image
 
## Logged in with projects
All your projects are listed with dates when you last edited them.
image

### DIFF
--- a/app/assets/stylesheets/editor/_project.scss
+++ b/app/assets/stylesheets/editor/_project.scss
@@ -1,0 +1,13 @@
+.project {
+  display: block;
+  padding: $margin / 8;
+  margin-bottom: $margin / 16;
+  border-radius: $border-radius / 2;
+  background: $bg-darker;
+  color: $text-color;
+  text-align: left;
+
+  &:hover {
+    background: darken($bg-dark, 7.5%);
+  }
+}
--- a/app/controllers/editor_controller.rb
+++ b/app/controllers/editor_controller.rb
@@ -22,6 +22,6 @@ class EditorController < ApplicationController
   private
 
   def current_user_projects(content_type)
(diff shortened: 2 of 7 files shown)


netux:
I like it, especially that project selector.

One thing I'd suggest is showing the Create project and Check out sample project buttons even if the user already has projects:
- I sometimes want to remember how to do something, and rather than use the wiki I'd prefer to quickly reference another project I know is using the thing I need correctly. The "Check out a sample project" button is a nice shortcut to get to such a project.
  - Your fishing mode doesn't use some of the editor's more obscure features (`@if`/`@for`/`@each`, `@contents`, `@translate`, etc.), so maybe for users with some projects already we could also link to a generic "showcase" project that just shows how these features can be used.
- The "Create project" button might get less usage, but it's still a nice shortcut to have if one is for some reason already in the empty state page. Of course, the label should be different if the user already has a project in their account.
- The text above the buttons can change to something like "Quick actions"

Mitcheljager:
@netux 
> , so maybe for users with some projects already we could also link to a generic "showcase" project

Good idea! For now I'll leave it as is though, as it might not showcase everything, it does showcase at least what the editor is for new users.

I've also added a "create new" button to the project display

image

PR #342, opened 2023-08-06
When compiling mixins in the Editor, avoid name overlap by replacing arguments with longer names first

netux:
Fixes a problem where given
```
@mixin foo(arg, arg2) {
  Mixin.arg2
}
```
`@include foo(a, b)` would render "a2" instead of "b"

### DIFF
--- a/spec/javascript/utils/compiler/mixins.test.js
+++ b/spec/javascript/utils/compiler/mixins.test.js
@@ -90,11 +90,11 @@ describe("mixins.js", () => {
       expect(disregardWhitespace(extractAndInsertMixins(input))).toBe(disregardWhitespace(expectedOutput))
     })
 
-    it("Should replace Mixin variables with given arguments", () => {
+    it("Should avoid replacing one Mixin variable with the value of another when the variable names overlap", () => {
       const input = `
-        @mixin testMixin(someArg, someOtherArg) {
+        @mixin testMixin(someArg, someArg2) {
           Mixin.someArg;
-          Some Action(Mixin.someOtherArg);
+          Some Action(Mixin.someArg2);
         }
         @include testMixin(One, Two)`
 
--- a/app/javascript/src/utils/compiler/mixins.js
+++ b/app/javascript/src/utils/compiler/mixins.js
@@ -95,11 +95,12 @@ export function extractAndInsertMixins(joinedItems) {
       replaceWith = replaceWith.replace("@contents;", contents || "")
     }
 
-    let paramIndex = 0
-    ;[... mixin.params].sort((p1, p2) => p2.key.length - p1.key.length).forEach(param => {
-      replaceWith = replaceWith.replaceAll("Mixin." + param.key, splitArguments[paramIndex]?.trim() || param.default)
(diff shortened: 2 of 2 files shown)


Mitcheljager (approved):
Ah good point! Clever solution!

Mitcheljager:
Tests are failing, it seems

netux:
>      TypeError: 0 is not a function
> 
>        96 |     }
>        97 |
>     >  98 |     let paramIndex = 0
>           |                       ^
>        99 |     ([... mixin.params]).sort((p1, p2) => p2.key.length - p1.key.length).forEach(param => {
>       100 |       replaceWith = replaceWith.replaceAll("Mixin." + param.key, splitArguments[paramIndex]?.trim() || param.default)
>       101 |       paramIndex++

Interesting, seems we are hitting a parser differential.

The reason I added parenthesis around the `[... mixin.params]` was because we don't use semicolons, but semicolons would prevent JavaScript from reading
```js
let whatever = 0
[whatever2]
```
as
```js
let whatever = 0[whatever2]
```

By adding parenthesis, that becomes
```js
let whatever = 0([whatever2])
```
which browsers and even VSCode take as invalid so they split that into multiple statements. Whatever Jest is using to execute the tests (presumably Node.js, so V8?) does not do that.

@Mitcheljager the solution is as simple as adding the semicolon
```js
let paramIndex = 0;
[... mixin.params].sort(/* ... */)
```
or if you really hate semicolons, we could go with storing the mixin params copy into a variable
```js
let paramIndex = 0
const mixinsParamsCopy = [... mixin.params]
mixinsParamsCopy.sort(/* ... */)
```

Which one would you prefer?

Mitcheljager:
@netux Typically in the no-semicolon world the convention would be to put the semicolon at the start of the line

```js
let paramIndex = 0
;([... mixin.params])....
```

netux:
Turns out `paramIndex` was an issue in the first place. Because we sort the parameters, `paramIndex` would go out-of-sync with the actual indices (thank the tests for catching that!).
Should be fixed now :)

spec/javascript/utils/compiler/mixins.test.js
```diff
@@ -90,11 +90,11 @@ describe("mixins.js", () => {
...
-    it("Should replace Mixin variables with given arguments", () => {
+    it("Should avoid replacing one Mixin variable with the value of another when the variable names overlap", () => {
       const input = `
-        @mixin testMixin(someArg, someOtherArg) {
+        @mixin testMixin(someArg, someArg2) {
           Mixin.someArg;
-          Some Action(Mixin.someOtherArg);
+          Some Action(Mixin.someArg2);
         }
         @include testMixin(One, Two)`
```
Mitcheljager, line 99: This should be a separate test, not replace the current one

netux: Actually, it turns out the test above this one was exactly the same. I'm just replacing the duplicate.

Mitcheljager: Ah fair enough

PR #344, opened 2023-08-11
Actuall check all ranges in `matchAllOutsideRanges()`

netux:
Fixes variable-likes inside strings being extracted (test included).

Also, expand range by one character in `findRangesOfStrings()` to account for the closing quote.


Mitcheljager (approved):
Nice fix!

PR #346, opened 2023-08-13
Replace `KeyboardEvent.key` → `KeyboardEvent.code` wherever possible

netux:
Turns out `.key` returns the text the user would have typed on a text input.

For example, if the user was pressing Shift + S, then `.key` would be `"S"`. If the user wasn't pressing Shift, then `.key` would be `"s"`.

This makes the Compile keybind work again.


Mitcheljager (approved):
Ahh, I hadn't noticed. good catch

PR #347, opened 2023-08-13
Rename Flashbang Explosion Sound effect to match Overwatch's changes

netux:
With this, he who shall not be named is no longer in named in the Workshop


Mitcheljager (approved):
Mcgone

PR #348, opened 2023-08-14
feat: Editor project backups

Mitcheljager:
This PR adds backups for editor projects. There is a new option under the "save" button, allowing you to create a backup. What this does is create a copy of your current project. It's not a direct copy, it's under a new model. Only the content and title gets copied. A project can have an unlimited amount of backups. You can load a backup and it will replace the current content with the backup. The project will not actually be replaced until you save, but ctrl+z'ing is impossible after replacing from the backup. The only way to restore is by refreshing. Ideally of course if you've pressed the restore from backup button, you meant to do it and there's no need to go back. We can rework that if that proves incorrect.

Additionally this PR refactors modals in some fashion. There now is a general Modal component as well as a modal store. Calling `modal.set(key)` will show the modal of that key. Calling `modal.close()`  will close any open modal. This is used for the new backups modal, as well as for the create project modal. I want to do this for all modals, but that will need to wait for a separate PR.

### Save dropdown
There now is a dropdown that is grouped with a save button. This opens a menu that shows your last save, allows you to create a backup, and allows you to view your backups.
image

### Backups modal
The backups modal shows all of your backups, in order from newest to oldest. 
| No backups | Some backups | Error state (not very pretty, but at least it's clear) |
| --- | --- | --- |
| image | image | image |

### Backup actions
The context menu for each backup allows you to either delete that backup or restore that backup. When restoring it replaces the current project content with the content of that backup. This includes translations.
image

---

backups

### DIFF
--- a/app/assets/stylesheets/editor/_project-backups.scss
+++ b/app/assets/stylesheets/editor/_project-backups.scss
@@ -1,0 +1,14 @@
+.project-backups {
+  display: flex;
+  flex-direction: column;
+  gap: $margin / 8;
+}
+
+.project-backup {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  padding: $margin / 8;
+  border-radius: $border-radius / 2;
+  background: $bg-darker;
+}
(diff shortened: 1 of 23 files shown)


Mitcheljager (commented):
Webhook test 👀

Mitcheljager (commented):
Webhook test #2 👀

PR #349, opened 2023-08-15
Allow for pressing Enter across some inputs in the Editor to execute their primary action

netux:
Introduces a `submittable` Svelte action which captures the Enter key on inputs and calls the `submit` event.
There is a special case for textareas, where using a modifier key + Enter will not trigger the event.

Also modifies some dialogs across the editor to `use:submittable`

### DIFF
--- a/app/javascript/src/components/actions/submittable.js
+++ b/app/javascript/src/components/actions/submittable.js
@@ -1,0 +1,26 @@
+
+export function submittable(node) {
+  const nodeIsInput = node.tagName === "INPUT"
+  const nodeIsTextarea = node.tagName === "TEXTAREA"
+
+  if (!(nodeIsInput || nodeIsTextarea)) {
+    throw new Error("submittable action is only meant to be used on inputs")
+  }
+
+  function keydown(event) {
+    if (event.code !== "Enter") return
+    if (nodeIsTextarea && (event.metaKey || event.altKey || event.shiftKey || event.ctrlKey)) return
+
+    event.preventDefault()
+
+    node.dispatchEvent(new CustomEvent("submit"))
+  }
+
+  window.addEventListener("keydown", keydown)
+
+  return {
+    destroy() {
+      window.removeEventListener("keydown", keydown)
+    }
+  }
+}
--- a/app/javascript/src/components/editor/ScriptImporter.svelte
+++ b/app/javascript/src/components/editor/ScriptImporter.svelte
@@ -2,14 +2,18 @@
   import { items } from "../../stores/editor"
   import { createNewItem } from "../../utils/editor"
   import { getSettings } from "../../utils/parse"
+  import { submittable } from "../actions/submittable"
   import { escapeable } from "../actions/escapeable"
   import { fade } from "svelte/transition"
 
   let active = false
   let replaceScript = false
   let value = ""
+  let disallowSubmit
 
   $: if (active) value = ""
+
+  $: disallowSubmit = !value
 
   function findSettings() {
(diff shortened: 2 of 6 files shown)


app/javascript/src/components/editor/ScriptImporter.svelte
```diff
@@ -64,15 +70,21 @@
     <div class="modal__content">
       Copy your snippet from inside of Overwatch and paste it in here to convert your current snippet to this project.
 
-      <textarea class="form-input form-textarea form-textarea--extra-small mt-1/4" bind:value />
+      <textarea
+        class="form-input form-textarea form-textarea--extra-small mt-1/4"
+        bind:value
+        use:submittable
+        on:submit={submit}
+      />
```
Mitcheljager, line 78: The closing tag should still be on the same line as the same attribute

```suggestion
        on:submit={submit} />
```

app/javascript/src/components/actions/submittable.js
```diff
@@ -0,0 +1,26 @@
...
+  const nodeIsInput = node.tagName === "INPUT"
+  const nodeIsTextarea = node.tagName === "TEXTAREA"
+
+  if (!(nodeIsInput || nodeIsTextarea)) {
+    throw new Error("submittable action is only meant to be used on inputs")
+  }
+
+  function keydown(event) {
+    if (event.code !== "Enter") return
+    if (nodeIsTextarea && (event.metaKey || event.altKey || event.shiftKey || event.ctrlKey)) return
```
Mitcheljager, line 12: I think this should happen for all inputs, not just textarea. If you are holding a modifier key your intent is most likely _not_ to submit

```suggestion
  function keydown(event) {
    if (event.code !== "Enter") return
    if (event.metaKey || event.altKey || event.shiftKey || event.ctrlKey) return
```

app/javascript/src/components/actions/submittable.js
```diff
@@ -0,0 +1,26 @@
+
+export function submittable(node) {
+  const nodeIsInput = node.tagName === "INPUT"
+  const nodeIsTextarea = node.tagName === "TEXTAREA"
+
+  if (!(nodeIsInput || nodeIsTextarea)) {
+    throw new Error("submittable action is only meant to be used on inputs")
+  }
```
Mitcheljager, line 8: I don't think think textarea should be included, the default behaviour of enter on textarea should still be new lines, as that is expected. If you don't know you will accidentally submit too early.
```suggestion

  if (!nodeIsInput) {
    throw new Error("submittable action is only meant to be used on inputs")
  }
```

Mitcheljager (changes requested):
Nice use of actions! I left some formatting suggestions, but nothing major

app/javascript/src/components/actions/submittable.js
netux: How about allow this in textareas but only when pressing Ctrl + Enter?

Mitcheljager: Sure, I like that!

app/javascript/src/components/editor/ScriptImporter.svelte
netux: I see a few cases of this. Might be worth investigating if eslint can detect that?
image

Mitcheljager: The ones that aren't false positives are all by you I believe. I think ESLint is able to detect this, but I'm not sure of the exact setting.

Mitcheljager (approved):
Tiny comment, but approving it either way

PR #351, opened 2023-08-16
refactor: Use svelte component for alerts

Mitcheljager:
The alerts were a bit of a mishmash, using javascript inside of svelte components. This made enough sense in the main website, but in the editor it's not the best experience.

This PR refactors it to use a svelte component. This component is still used in the main website as well, so it's not just for the editor. To be able to use this from any javascript, instead of setting some store or something, it fires a custom event that is captured by the alerts component. It now uses svelte transitions as well, which is neat.

It still is a bit of a mishmash in the sense that there is also the rails flash, which use the exact same structure and classes, but is just flat HTML. Odd, but that's the way of things.

### DIFF
--- a/app/views/application/_flash.html.erb
+++ b/app/views/application/_flash.html.erb
@@ -7,3 +7,5 @@
     </div>
   <% end %>
 </div>
+
+<%= svelte_component "Alerts" %>
--- a/app/views/application/_flash.html.erb
+++ b/app/views/application/_flash.html.erb
@@ -1,11 +1,1 @@
-<div class="alerts" data-role="alerts">
-  <% flash.each do |type, message| %>
-    <% next if type == "alert" %>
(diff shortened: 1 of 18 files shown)


app/views/application/_flash.html.erb
```diff
@@ -7,3 +7,5 @@
     </div>
   <% end %>
 </div>
+
+<%= svelte_component "Alerts" %>
```
netux, line 11: What if the `.alert` element in `Alerts.svelte` was factored out into e.g. `Alert.svelte`, and we used `svelte_component "Alert"` to render the alerts from flash?

Mitcheljager: Hmm actually, what you just said made me realise we can render the flash as json as a prop, and pass it to the svelte component.

Mitcheljager:
@CactusPuppy I did a dirty and disabled a bunch of tests because they were failing. I couldn't figure out how to detect json content rather than plain text on the page. If you have any pointer that would be much appreciated!

PR #352, opened 2023-08-16
Fix some issues related to variable extraction and compilation

netux:
- Restrict variable regex to only letters, numbers, or underscore
- Avoid giving default variable names a new index, causing an error on import (details in d31c8e6d1c4e361a76719b2e44ffcea2bcf26c45)

Both changes include tests.


Mitcheljager (approved):
Nice! Some of it is a bit difficult to read, but that's ok. Nice tests too!

PR #353, opened 2023-08-18
Add mobile dropdown menu with action buttons for the Editor mobile view

netux:
- Always show the Settings icon, even on mobile
- Never show the Shortcuts icon on mobile
- Include Translations, Import Script, and Compile buttons to the mobile dropdown
  - Add a new `inline` mode to the Compile button

While the mobile view shouldn't get much love, I was triggering it by having DevTools open on the side.

This PR allows mobile (or small screen) users to use these action buttons.

Open mobile dropdown menu


app/javascript/src/components/editor/EditorActions.svelte
```diff
@@ -17,21 +30,57 @@
...
       </button>
     {/if}
+  {/if}
 
+  {#if !$isMobile}
     <Compiler />
   {/if}
 
+  {#if $isMobile}
+  <div class="dropdown settings" bind:this={mobileDropdown}>
```
Mitcheljager, line 56: An indent is missing

Mitcheljager (changes requested):
Overall a very nice change! I did leave some comments regarding formatting, mostly in the name of keeping things easy to maintain.

netux (commented):
~~style(editor actions): add missing indent~~
Wait, that is not the commit input 🤦‍♂️

app/javascript/src/components/editor/Compiler.svelte
```diff
@@ -41,17 +43,25 @@
...
   {/if}
 
-  {#if copied}
+  {#if copied && !inline}
     <div transition:fly={{ y: 5, duration: 150 }} class="tooltip__content bg-primary text-pure-white block">
       Copied to clipboard
     </div>
   {/if}
-</button>
+</svelte:element>
```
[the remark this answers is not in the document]
netux: The reason I had to do the "Compiled and copied!" text was because the tooltip was inserted inside the dropdown. However, I think we could probably move this alert to be an actual alert (once #351 is merged).

I will look into the `.empty-button` class, thanks for the data!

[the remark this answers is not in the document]
Mitcheljager: With this PR (https://github.com/Mitcheljager/workshop.codes/pull/356) merged you will no longer need to worry about that, it can simply be a button.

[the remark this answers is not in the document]
netux: Seems like it can simply be a button right now too. I noticed `.empty-button` changed the font family of the button, so I had to remove it. Instead, `button.dropdown__item` applies similar styles.

app/assets/stylesheets/elements/_dropdown.scss
```diff
@@ -124,6 +124,11 @@
     background-size: 70%;
   }
 }
+button.dropdown__item {
+  width: 100%;
+  background-color: transparent;
+  border: none;
+}
```
Mitcheljager, line 131: Styling specific to an element is annoying because it's not predictable, I will merge my other PR first, and you should be able to remove this.

app/javascript/src/components/editor/Compiler.svelte
```diff
@@ -24,11 +32,9 @@
   }
 
   function copyToClipboard(value) {
-    copied = true
-
     copyValueToClipboard(value)
 
-    setTimeout(() => copied = false, 1000)
+    addAlert("Copied compiled snippet to clipboard")
```
Mitcheljager, line 37: I don't think it should be an alert. It is much nicer inline since that's where your eyes already are

app/javascript/src/components/editor/Compiler.svelte
```diff
@@ -41,17 +47,13 @@
...
     Compiling...
   {:else}
     Compile
   {/if}
-
-  {#if copied}
-    <div transition:fly={{ y: 5, duration: 150 }} class="tooltip__content bg-primary text-pure-white block">
-      Copied to clipboard
-    </div>
-  {/if}
```
Mitcheljager, line 55: We should keep this as it was

app/javascript/src/components/editor/EditorActions.svelte
```diff
@@ -6,8 +6,23 @@
   import Settings from "./Settings.svelte"
   import Shortcuts from "./Shortcuts.svelte"
   import Save from "./Save.svelte"
+  import ThreeDotMenu from "../icon/ThreeDotMenu.svelte"
+
+  let mobileDropdown
+  let showMobileDropdown = false
+
+  $: if (!$isMobile) showMobileDropdown = false
```
Mitcheljager, line 14: Is the idea of this to set it to false if you change from mobile to desktop while the menu is open?

app/javascript/src/components/editor/Compiler.svelte
```diff
@@ -1,10 +1,18 @@
...
+
   let compiling = false
-  let copied = false
+  let classes
+
+  $: classes = [
+    "tooltip",
+    $$restProps["class"],
+    !inline && "button button--secondary button--square"
+  ].filter((c) => !!c).join(" ")
```
Mitcheljager, line 15: This is kind of hard to read, I feel it would be better to just do it inline on the element itself.

Mitcheljager (changes requested):

app/javascript/src/components/editor/EditorActions.svelte
netux: Indeed. Just a small thing to close the dropdown when it won't be shown anyway, so if the user changes from desktop to mobile again the dropdown isn't still open.

app/javascript/src/components/editor/Compiler.svelte
netux: I have reverted back to the tooltip.
Are you still not okay with the amount of `if`s needed for the button's text? Splitting the button into an inline and not inline version is a little awkward due to the shared logic on click.

Mitcheljager: Actually I think by now it's totally ok as is, before it was mostly with the extra logic for it being a button or a span, that it felt like it could be more readable if it was split up. But as it is now I think it's definitely readable enough, nicely done.

app/javascript/src/components/editor/Compiler.svelte
```diff
@@ -41,14 +43,20 @@
 
 <svelte:window on:keydown={keydown} />
 
-<button class="button button--secondary button--square tooltip" on:click={doCompile}>
-  {#if compiling}
-    Compiling...
+<button
+  class={"tooltip " + ($$restProps["class"] ?? "") + " " + (inline ? "empty-button" : "button button--secondary button--square")}
```
Mitcheljager, line 47: You can use `{ ... }` inside the property to make it a little more readable

```suggestion
  class="tooltip {$$restProps["class"]} {inline ? "empty-button" : "button button--secondary button--square"}"
```

Mitcheljager (approved):
Nice!

PR #354, opened 2023-08-18
Add "Mixin.whatever" as a valid iterable for `findEachLoopsWithInvalidIterables` the Editor linter rule

netux:
Fixes diagnostic errors in the following code, because "Mixin.Buttons" would not have been a valid variable iterable
```haskell
@mixin TempAllowAndPressButtons(Player, Buttons) {
	@each (button in Mixin.Buttons) {
		Allow Button(Mixin.Player, Mixin.Button);
		Press Button(Mixin.Player, Mixin.Button);    
		Disallow Button(Mixin.Player, Mixin.Button);            
	}
}

@include TempAllowAndPressButtons(Event Player, [Button(Primary Fire), Button(Jump)])
```


Mitcheljager (approved):
Ah I hadn't thought of that, nice and easy fix!

PR #357, opened 2023-08-20
Add `outsideClick` Svelte action

netux:
This is another pattern we can extract away into an Svelte action.
Though since it is pretty much only used for dropdowns, maybe we should just have a dropdown component.

More details in the commit messages.

### DIFF
--- a/app/javascript/src/components/editor/EditorItemCreate.svelte
+++ b/app/javascript/src/components/editor/EditorItemCreate.svelte
@@ -20,6 +20,13 @@
     if (type == "item") $currentItem = newItem
   }
 
+  function withCloseDropdown(fn) {
+    return (... args) => {
+      fn(... args)
+      active = false
+    }
+  }
+
   function toggle() {
     active = !active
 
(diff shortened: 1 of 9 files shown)


app/javascript/src/components/editor/EditorItemCreate.svelte
```diff
@@ -24,30 +20,35 @@
     if (type == "item") $currentItem = newItem
   }
 
+  function withCloseDropdown(fn) {
+    return (... args) => {
+      fn(... args)
+      active = false
+    }
+  }
```
Mitcheljager, line 28: This is kind of ugly, but I understand the purpose. A different option might be to simple add `active = false` to both `createTemplate` and `createItem`

Mitcheljager (approved):
Nice! I left one question, but I think it's also ok as is, so I will leave that up to you.

I definitely want to add a dropdown compontent. Initially I didn't add one because I was in a hurry to create to editor, but by now we do need one. Basically the same story as with the modals. But for now this is a good refactor, and the action is great addition either way. I will take care of create the component at a later time.

app/javascript/src/components/editor/EditorItemCreate.svelte
netux: Either option looked ugly to me. But this one was more explicit, which is why I went with it.

Mitcheljager: The reason I say this is ugly is because it takes more time to read and understand. While the other option is perhaps more repetition, at least you're able to understand it in an instant

PR #362, opened 2023-09-04
Remove any [linemarkers] in @each array items

netux:
Fixes an issue where an array iterable spanning multiple lines could confuse the compiler when run with linemarkers inserted.

I'm not very happy with having to account for the linemarkers in the compiler directly, I'd much rather change the line finder and compiler so this metadata isn't mixed with the actual code. But for now this will do.

##### Example problematic snippet

```
[linemarker]d3ae10|1[/linemarker]@each (pair, i in [
[linemarker]d3ae10|2[/linemarker]	// Name, Value
[linemarker]d3ae10|3[/linemarker]	["True", True]
[linemarker]d3ae10|4[/linemarker]]) {
[linemarker]d3ae10|5[/linemarker]	@each (thing, j in Each.pair) {
[linemarker]d3ae10|6[/linemarker]		Each.thing (Each.i);
[linemarker]d3ae10|7[/linemarker]	}
[linemarker]d3ae10|8[/linemarker]}
[linemarker]d3ae10|9[/linemarker]
```

this would've previously compiled to

```
  ["True", True]
[linemarker]d3ae10|4[/linemarker (0);
```

but now it correctly compiles to

```
  "True" (0);

  True (0);
```

### DIFF
--- a/app/javascript/src/utils/compiler/each.js
+++ b/app/javascript/src/utils/compiler/each.js
@@ -97,4 +97,7 @@ export function parseArrayValues(input) {
   }
 
   return result
+    // HACK: line finder inserts [linemarker]s on the input, which may confuse @each
+    // into thinking they are nested arrays.
+    .map((item) => item.replace(/\s*\[linemarker\].*?\[\/linemarker\]\s*/g, ""))
 }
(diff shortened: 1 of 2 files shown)


Mitcheljager (approved):
Makes sense. Perhaps we can change the line marker to something that isn't quite as common.

netux:
> Makes sense. Perhaps we can change the line marker to something that isn't quite as common.

No, that would just eventually move the problem somewhere else.

PR #363, opened 2023-09-15
feat: Use localStorage as fall back for saving projects

Mitcheljager:
We've had some reports of projects seemingly not saving after coming back to the editor. This PR doesn't fix that necessarily, but it adds a fall back. The project will now also be saved to your local storage, along with a timestamp. When fetching a project your local storage is also fetched. If the timestamp of the local storage project is newer than that of the fetched project, something may have gone wrong previously. In this case we use the content of the local storage project instead.

I thought about adding some alert along the lines of "Your local save is newer than the save on the server, do you want to use this instead?", similar to steam back ups saves perhaps, but it felt like this was too convoluted and most people won't have a clue what that means. So I decided to keep it stealthy and assume that the latest save is always the desired save.

This PR also fixes a missing import, but that's unrelated.

### DIFF
--- a/app/javascript/src/components/editor/Modals/CreateProjectModal.svelte
+++ b/app/javascript/src/components/editor/Modals/CreateProjectModal.svelte
@@ -1,7 +1,7 @@
 <script>
   import Modal from "./Modal.svelte"
   import { currentProject, isSignedIn, modal } from "../../../stores/editor"
-  import { createProject, renameCurrentProject } from "../../../utils/project"
+  import { createProject, renameCurrentProject, setUrl } from "../../../utils/project"
   import { submittable } from "../../actions/submittable"
 
   let loading
--- a/app/javascript/src/utils/project.js
+++ b/app/javascript/src/utils/project.js
@@ -58,6 +58,7 @@ export async function fetchProject(uuid) {
       // This is a fallback and should not be the norm.
       if (localProject && new Date(parsedData.updated_at) < new Date(localProject.updated_at)) {
         parsedData.content = localProject.content
+        addAlert("We recovered a version of your project that wasn't fully saved.")
       }
 
       updateProject(parsedData.uuid, {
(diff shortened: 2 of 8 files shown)


CactusPuppy:
I would agree that the latest timestamp save is probably the best option to load, but maybe something along the lines of "We recovered a version of your project that wasn't fully saved." would help indicate that something is a bit off behind the scenes?

app/javascript/src/utils/project.js
```diff
@@ -44,12 +44,22 @@ export function createDemoProject(title) {
...
       if (!data) throw Error("No results")
 
       const parsedData = JSON.parse(data)
 
+      // If the project in localStorage is newer than the project from the API
+      // something may have gone wrong when saving to the API, but the localStorage
+      // was still saved as expected. In this case we use the data from localStorage.
+      // This is a fallback and should not be the norm.
+      if (localProject && new Date(parsedData.updated_at) < new Date(localProject.updated_at)) {
+        parsedData.content = localProject.content
```
CactusPuppy, line 60: I wonder if we could send an event over to Bugsnag here or something so we can track how often this incident happens?

app/javascript/src/components/editor/Modals/CreateProjectModal.svelte
```diff
@@ -1,7 +1,7 @@
 <script>
   import Modal from "./Modal.svelte"
   import { currentProject, isSignedIn, modal } from "../../../stores/editor"
-  import { createProject, renameCurrentProject } from "../../../utils/project"
+  import { createProject, renameCurrentProject, setUrl } from "../../../utils/project"
```
CactusPuppy, line 4: Pretty sure this doesn't need to be here?

Mitcheljager: It does, there was a missing import

app/javascript/src/utils/project.js
Mitcheljager: Good idea! We don't have bugsnag set up for javascript currently, but I will have a look

Mitcheljager: Turns out it was fairly easy, their javascript package is pretty nice! I will add this to other parts of the editor as well, but I will save that for another PR

PR #366, opened 2023-09-21
Improve editor detection of variables in the context of numbers

netux:
- Avoid detecting variables from numbers without
 a whole number part (e.g. ".1234")
- Handle beginning of text in the regex (e.g. don't detect variables when the input is literally "0.abc" or "D.abc")


Mitcheljager (approved):
Nice!

PR #367, opened 2023-09-23
feat: Video attachments

Mitcheljager:
This PR adds video attachments to posts and wiki articles. Similar to images you can upload a video using the markdown editor and it will automatically insert a video tag. Videos are limited to mp4 and 50MB file size. We could easily add other video formats if we so desire, but I decided to keep it simple for now.

When files are uploaded to Digital Ocean they are automatically set to private, this can be changed by setting `public: true` in `storage.yml`, except that this only sets it public when create a variant, for reasons I'm not sure of. The original file would still be private until that variant was made. This was fine for images, as all images have variants anyway. For videos we don't want to create variants, the original is all we need, but this meant it was never set to public. I ended up solving this by patching the  `@rails/activestorage` npm package by including the header `x-amz-acl=public-read`. I couldn't do this anywhere as as it's DirectStorage that handles the uploading. This means all files are uploaded as public, which is exactly what we need! It's not the cleanest solution of all time, but after hours of debugging this is the best I got.

---

Videos can be uploaded using the button in the editor, they can't be pasted like images can.
| Wiki | Posts |
| --- | --- |
| image | image |

When the video is uploaded it will add a new tag to the description with the video url 
image

The video is included in the post as a regular video tag. There's no special player.
image


netux:
Since you had to patch `@rails/activestorage`, have you considered creating a bug report on [rails/rails](https://github.com/rails/rails/issues)? Or even better, solve the issue for them if you have the time :D

Mitcheljager:
@netux I don't think it's quite that easy, as most people don't want public files. This is also specific to S3, and I can't be bother making it not specific :D

app/helpers/content_helper.rb
```diff
@@ -91,6 +92,12 @@ def markdown_youtube(text)
     end
   end
 
+  def markdown_video(text)
+    text.gsub /\[video\s+(.*?)\]/ do
```
netux, line 96: This matches "[video ]"

app/helpers/content_helper.rb
```diff
@@ -91,6 +92,12 @@
...
+      "<div class='video'>#{ActionController::Base.helpers.video_tag($1, controls: true)}</div>"
+    end
```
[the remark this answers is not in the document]
netux: Keyword "may". Tried with Regex101 and it consistently takes 0ms to match:
image

https://regex101.com/r/BbDYJj/1

netux:
> @netux I don't think it's quite that easy, as most people don't want public files. This is also specific to S3, and I can't be bother making it not specific :D

Then just a ticket to request options on whatever upload function used here (for the S3-specific package, or not) that allows for setting the privacy settings of the file on upload? You said you can only set the privacy when generating variants, which seems like a bug to me when you can't do it for anything else.

Admittedly though, I'm unfamiliar with this rails activestorage and with uploading things to S3, so maybe in reality it doesn't work quite as I imagine.

app/javascript/src/inscryb-mde.js
```diff
@@ -296,6 +301,44 @@ class InitialiseInscrybeMDE {
...
+      labelElement.innerText = "Upload video"
+      labelElement.classList.add(...labelClasslist)
+
+      const inputElement = document.createElement("input")
+      inputElement.type = "file"
+      inputElement.id = randomId
+      inputElement.accept = "video/mp4"
+      inputElement.classList.add("hidden-field")
+
+      inputElement.addEventListener("change", () => { new InscrybeInsertVideo(event, this.codemirror).input() })
```
netux, line 329: VSCode is warning me that `event` is deprecated.

```suggestion
      inputElement.addEventListener("change", (event) => { new InscrybeInsertVideo(event, this.codemirror).input() })
```

Technically this works because of funny Browser behaviour (see [`window.event`](https://developer.mozilla.org/en-US/docs/Web/API/Window/event)).

app/javascript/src/inscryb-mde.js
```diff
@@ -296,6 +301,44 @@ class InitialiseInscrybeMDE {
...
+    button.classList.toggle("dropdown-open")
+
+    if (button.classList.contains("dropdown-open")) {
+      const dropdownElement = document.createElement("div")
+      dropdownElement.classList.add("editor-dropdown")
+
+      const textElement = document.createElement("small")
+      textElement.innerText = "Upload a video. Limited to mp4 filetype and 50mb filesize."
+
+      const randomId = Math.random().toString().substr(2, 8)
```
netux, line 316: Why use a `randomId` in the first place? I don't think there can be multiple of these dropdowns open, right?

It could just be a static, but very specific, ID. Something like "inscrybe-upload-video-file-input".

netux (approved):
LGTM! I appreciate that it works flawlessly on my local setup too :)

app/helpers/content_helper.rb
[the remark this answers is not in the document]
Mitcheljager: I will fix it though, might as well. The input is should always be the same anyway

app/javascript/src/inscryb-mde.js
Mitcheljager: Good question, I don't quite remember. This was copied from the image uploader

Mitcheljager: Ah I remember now, it's because you can have the markdown editor open twice on the same page, for example 1 for the description and 1 for the revision description.

app/helpers/content_helper.rb
Mitcheljager: I ended up changing it to `/\[video\s(https?:\/\/\S+)\]/`, this ensures it's a url as well since [video abc] would have resulted in an error.

PR #368, opened 2023-09-27
feat: Replace category select

Mitcheljager:
This PR replaces the multi select box for categories with a svelte component. The multi select is unpredictable and annoying to use. By using regular checkboxes we make the whole UX make clear, as you don't need to hold an additional button to select more categories. The checkboxes automatically limit themselves to 3 items, making the validation much simpler as well. Even if you didn't read can you only select 3, you will be forced to. The old validation is still in place in case someone tries to game the system.

category-select

### DIFF
--- a/app/javascript/src/components/form/LimitedCheckboxes.svelte
+++ b/app/javascript/src/components/form/LimitedCheckboxes.svelte
@@ -1,0 +1,39 @@
+<script>
+  export let name = "checkboxes"
+  export let categories = []
+  export let selected = []
+  export let limit = 3
+
+  /**
+   * Add the selected option to the list of selected options.
+   * If you select more than the given limit the first item will be replaced with the new one.
+   * E.g if you selected 1,2,3 and now add 4, the next order will be 2,3,4
+   * @param {boolean} checked The state of event checkbox
+   * @param {string} value The given value to be added to the selected array
+   */
+  function bind(checked, value) {
+    if (!checked) {
+      selected = selected?.filter(o => o !== value)
+      return
+    }
+
+    if (selected.length >= limit) selected = [...selected.slice(1), value]
+    else selected = [...selected, value]
+  }
+</script>
+
+<div class="well well--scrollable block br-1/2">
+  {#each categories as [label, value]}
+    <div class="checkbox">
+      <input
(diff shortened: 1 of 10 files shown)


netux:
Instead of the first-in first-out approach to limit the amount of categories you can select, how about disabling all checkboxes when the user has selected all 3 categories?

Mitcheljager:
@netux I thought about that, but that felt more annoying

netux:
How about giving the checkboxes label's `width: 100%`?
before
after

That way users can click anywhere inside the block to select an option.

Mitcheljager:
@netux Good idea!

app/javascript/src/components/form/LimitedCheckboxes.svelte
```diff
@@ -0,0 +1,54 @@
+<script>
+  export let name = "checkboxes"
+  export let categories = []
```
netux, line 3: Since this component is meant to be generic, this could probably be renamed to something like `items`

netux (approved):
Frontend looks good to me.

app/javascript/src/components/form/LimitedCheckboxes.svelte
Mitcheljager: Ah for sure!

netux:
One issue: deselecting all categories doesn't seem to work.

Mitcheljager:
@netux This was an issue before as well, and it's an underlying issue with how Rails handles empty inputs. On the positive side, that would be invalid anyway

netux:
> that would be invalid anyway

How is it invalid if you are never forced to set a category?


> it's an underlying issue with how Rails handles empty inputs

Hmm. Well, now that there is a \<LimitedCheckboxes> component, why not have a hidden checkbox that is checked when all checkboxes are disabled, and handling that on the backend?

Mitcheljager:
@netux That is exactly how it works, normally when you have a single checkbox a hidden field is added with a value of 0 that is used when the actual value is unchecked. But when it's an array that's can't be done, because then the value is part of the array. The same happens when you deselect all heroes or maps for example.
I don't want to fix that in this PR though, as it's not that easy.

PR #371, opened 2023-10-05
Detect variables in the editor after a Workshop value with arguments/a closing parenthesis

netux:
E.g. `Ray Cast Hit Player(...).someVariable`.


Mitcheljager (approved):
That's certainly much easier than expected

PR #376, opened 2023-10-08
Improve Editor variable excluding logic to better match what the game wants

netux:
- Fixed an off-by-one bug that saved some variables from being excluded when they shouldn't
- Fixed an issue where variables past Z were being excluded

Also added JSDocs explaining how the logic should work, and improved tests by splitting them up into smaller testable chunks

### DIFF
--- a/app/javascript/src/utils/compiler/variables.js
+++ b/app/javascript/src/utils/compiler/variables.js
@@ -43,6 +43,25 @@ export function getDefaultVariableNameIndex(name) {
   return index - 1
 }
 
+/**
+ * Exclude variables that would already be defined by default.
+ *
+ * For example, the following would fail because Overwatch already declares "B" at index 1 by default:
+ * ```
+ *   global:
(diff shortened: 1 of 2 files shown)


app/javascript/src/utils/compiler/variables.js
```diff
@@ -43,14 +43,61 @@ export function getDefaultVariableNameIndex(name) {
...
+ *     26: AZ // this is also fine, because we rename index 51 (what would be AZ)
+ *     // ...
+ *     50: AY
+ *     51: someNameThatIsNotJustAZ
+ *     52: BA
+ * ```
+ *
+ * @param {string[]} variables A list of variables
+ * @returns {string[]} The list of variables without
+ */
```
Mitcheljager, line 92: The longest comment ever, at least it makes it clear :D

Mitcheljager (approved):
Nice job!

PR #378, opened 2023-10-10
feat: Ollie login form

Mitcheljager:
With this PR Ollie will accompany you while you enter your log in details. It's a little bit silly, but it's fun! The code isn't the cleanest ever, but I wanted to keep it fairly self contained, without having to add more data attributes to other elements, and without having to keep some state.

ollie-form


netux:
So the pupils are above the face? 
image

How about making the eyes transparent and putting the pupils behind the face?

Mitcheljager:
@netux Good idea!

app/javascript/src/ollie-form.js
```diff
@@ -0,0 +1,58 @@
...
+  rememberMeInput.removeAndAddEventListener("input", () => setOllieBody(element, "base"))
+}
+
+function setPupilPosition({ target }, element) {
+  const max = 32
+  const valueLength = Math.min(max, Math.max(0, target.value.length))
+  const xOffset = -1 + (valueLength / (max / 2)) * 2
+
+  const pupils = getOlliePart(element, "pupils")
+  pupils.style.transform = `translateX(${ xOffset }px) translateY(1px)`
```
netux, line 26: Maybe using percentages may give smaller steps. From my testing, 14% seems to be a good max percentage.

app/javascript/src/ollie-form.js
```diff
@@ -56,3 +58,24 @@ function setOllieBody(element, variant = "base") {
   armLeft.classList.toggle("out-of-view", variant !== "eyes-closed")
   armRight.classList.toggle("out-of-view", variant !== "eyes-closed")
 }
+
+function loadImagesSimultaniously() {
+  const element = document.querySelector("[data-role='ollie-image-holder']")
+  const images = element.querySelectorAll("img")
+
+  console.log(images)
```
netux, line 66:
```suggestion
```

Mitcheljager:
@netux I've made the adjustments you suggested, along with some more. Ollie will now only show when all images are loaded, to prevent the pupils from showing before anything else. Additionally I've made the CSS a bit more flexible. All sizing is now based off a single variable, so we can change the size super easily if we wanted to.

ollie-loading
ollie-form-big

PR #381, opened 2023-10-13
feat(editor): Remember scroll state between different items

Mitcheljager:
With this PR your scroll state will be remembered when navigating between different items. So if you scroll 100 pixels on file A, go to file B, and back to file A, file A will still be scroll 100 pixels.

The event is bound using a codemirror plugin. Unfortunately this event is also fired when the state updates, so when changing between items. This messes with the scroll position, as it would set it to a seemingly random scroll position. To prevent this from happening I added a variable that is toggled when changing between two items. If this variable is true the scroll position isn't saved. This is wrapped in a `setTimeout` with no time set to wait for the next render tick, so as to wait for the view update to complete.

scroll-state

### DIFF
--- a/app/javascript/src/components/editor/CodeMirror.svelte
+++ b/app/javascript/src/components/editor/CodeMirror.svelte
@@ -1,5 +1,5 @@
 <script>
-  import { onDestroy, onMount, createEventDispatcher } from "svelte"
+  import { onDestroy, onMount, createEventDispatcher, tick } from "svelte"
   import { basicSetup } from "codemirror"
   import { EditorView, keymap } from "@codemirror/view"
   import { EditorState, EditorSelection } from "@codemirror/state"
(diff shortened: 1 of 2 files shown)


app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -52,7 +55,21 @@
...
+  }
+
+  /**
+   * This is fired after updateEditorState, after all required bits are set.
+   * Part of this is wrapped in an empty setTimeout to wait for the view state to update.
+   */
+  function onStateUpdateEnd() {
     view.setState($editorStates[currentId])
+
+    setTimeout(() => {
```
netux, line 69: Maybe [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)  could work here?

Mitcheljager: Sure! Ultimately it's the same result, but requestAnimationFrame is a bit more explicit I suppose

netux: I was hoping it'd remove that frame where the scrollbar shifts after changing back to the file, as shown on your gif. I guess it did not.

Mitcheljager: Oh you know what, it did do that!

scroll-state-2

PR #391, opened 2023-12-18
prevent error when archive controller is provided a post code that doesn't exist

### LINKED ISSUE #389: [Bug] `/archive/[code that doesn't exist]` is not properly handled
Attempting to access the `/archive/[code]` when the code is not a valid code on Workshop.codes results in an HTTP status code of 500 instead of a 404.
image

CactusPuppy:
Fixes #389


Mitcheljager (commented):
Nice, thanks!

PR #392, opened 2023-12-19
feat(editor): Parameter Objects

Mitcheljager:
When working with actions or values with many parameters it can be quite a hassle. Either you only end up changing one or two of the values, or you end up with many values that make it difficult to see what is what. To address this this PR adds parameter objects. In terms of javascript you can look at it like a function `someFunction(someValue, someOtherValue)` vs `someFunction({ param1: someValue, param2: someOtherValue })`. The syntax I implemented here is very similar to that.

For example in this action all I'm doing is changing the position:
```js
Create In-World Text(All Players(All Teams), Custom String("Some string"), Vector(9, 9, 9), 1, Clip Against Surfaces, Visible To Position And String, Color(white), Default Visibility);
```
With parameter objects we can write it like this instead:
```js
Create In-World Text({ Position: Vector(9, 9, 9) });
```
And you could add more like so:
```js
Create In-World Text({ Position: Vector(9, 9, 9), text: Custom String("Some custom string") });
```
It can also help make your code much more readable, even when using all values:
```js
Create In-World Text({
  Players: All Players(All Teams),
  Text: Custom String("Some string"),
  Position: Vector(9, 9, 9),
  Scale: 1,
  Clipping: Clip Against Surfaces,
  Reevaluation: Visible To Position And String,
  Text Color: Color(white),
  Spectators: Default Visibility
});
```
Objects can be nested:
```js
Create In-World Text({ Position: Vector({ Y: 10 }) });
```

I've also added editor warnings, warning you about incorrectly used keys.

image

This is all optional of course. In the future I want to add a toggle to make it so autocomplete automatically inserts these objects rather than their default values, but let's see what the response and demand is first.

### DIFF
--- a/app/javascript/src/utils/compiler/parameterObjects.js
+++ b/app/javascript/src/utils/compiler/parameterObjects.js
@@ -20,15 +20,13 @@ export function evaluateParameterObjects(joinedItems) {
 }
 
 export function getFirstParameterObject(content) {
-  const regex = /\(\s*{/
+  const regex = /[a-z]\(\s*{/
   const match = content.match(regex)
 
   if (!match) return null
 
-  const phrase = getPhraseFromIndex(content, match.index - 1)
+  const phrase = getPhraseFromIndex(content, match.index)
   const completion = get(completionsMap).find(item => item.args_length && item.label === phrase)
-
-  if (!completion) return null
 
   const end = getClosingBracket(content, "{", "}", match.index)
   const string = content.slice(match.index + match[0].length, end).trim()
@@ -37,13 +35,24 @@ export function getFirstParameterObject(content) {
 
   splitParameters.forEach(item => {
     const [key, value] = item.split(/:(.+)/)
-    given[key.trim()] = value.trim()
+    given[key.trim()] = (value || "").trim()
   })
 
-  return {
-    start: match.index + 1,
+  const result = {
+    start: match.index + match[0].indexOf("{"),
     end,
-    given,
+    given
+  }
+
+  // Return a false object to replace contents of unfound phrase
+  if (!completion) return {
+    ...result,
+    phraseParameters: [],
+    phraseDefaults: []
+  }
+
+  return {
+    ...result,
     phraseParameters: completion.detail_full.split(", "),
     phraseDefaults: completion.parameterDefaults
   }
--- a/app/javascript/src/utils/compiler/parameterObjects.js
+++ b/app/javascript/src/utils/compiler/parameterObjects.js
@@ -53,7 +53,7 @@ export function getFirstParameterObject(content) {
 
   return {
     ...result,
-    phraseParameters: completion.detail_full.split(", "),
+    phraseParameters: completion.parameter_keys,
     phraseDefaults: completion.parameter_defaults
   }
 }
--- a/app/javascript/src/utils/compiler/parameterObjects.js
+++ b/app/javascript/src/utils/compiler/parameterObjects.js
@@ -20,7 +20,7 @@ export function evaluateParameterObjects(joinedItems) {
 }
 
 export function getFirstParameterObject(content) {
-  const regex = /[a-z]\(\s*{/
+  const regex = /[a-z]\s*\(\s*{/
   const match = content.match(regex)
 
   if (!match) return null
@@ -52,7 +52,7 @@ export function getFirstParameterObject(content) {
   return {
     ...result,
     phraseParameters: completion.detail_full.split(", "),
-    phraseDefaults: completion.parameterDefaults
+    phraseDefaults: completion.parameter_defaults
   }
 }
 
--- a/app/javascript/src/utils/compiler/parameterObjects.js
+++ b/app/javascript/src/utils/compiler/parameterObjects.js
@@ -5,12 +5,18 @@ import { get } from "svelte/store"
 export function evaluateParameterObjects(joinedItems) {
   let moreAvailableObjects = true
   let safety = 0
+  let startFromIndex = 0
 
   while (moreAvailableObjects) {
-    const parameterObject = getFirstParameterObject(joinedItems)
+    const parameterObject = getFirstParameterObject(joinedItems, startFromIndex)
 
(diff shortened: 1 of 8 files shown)


netux:
Interesting idea. I'll try and take a look at the implementation later.

But some early feedback:
- Since this is a language addition, I'm wondering if it should have a `@` somewhere. But I don't know where 🤔
  - Maybe as some sort of [`@pragma` directive](https://www.geeksforgeeks.org/pragma-directive-in-c-c/) to turn it on and off?
- I'm also not super fond of the syntax in general. I like how easy having to specify `{}`s around the functions must make it to detect them in the parser, but coming from Python I also like how [simple and flexible their keyword arguments are](https://docs.python.org/3/tutorial/controlflow.html#keyword-arguments).
  - One thing this is missing by having to wrap all arguments in a JSON-like structure is the mixing of positional and keyword arguments. I'd love to be able to do something like `WaitUntil(Is Alive(Host Player), timeout=15)` or `Set Status(Event Player, assiter=Null, Phased Out, duration=99999)`.

Mitcheljager:
>Since this is a language addition, I'm wondering if it should have a @ somewhere. But I don't know where 
I had the same thought, but I couldn't come up with anything that didn't feel forced. At least the bracket format is also unique to the editor (except for in custom strings).

> I'm also not super fond of the syntax in general. I like how easy having to specify {}s around the functions must make it to detect them in the parser, but coming from Python I also like how [simple and flexible their keyword arguments are](https://docs.python.org/3/tutorial/controlflow.html#keyword-arguments).
I can't say I'm a super big fan of the Python syntax. It's flexible perhaps, but it reads a quite confusing and as far as I can tell you still need to order things correctly. I'm a bit biased of course, since I'm familiar with the JS format but not the Python one

netux:
> At least the bracket format is also unique to the editor

For now*. Though any update to the workshop snippet syntax is highly unlikely at this point, so I guess this is fine.

> as far as I can tell you still need to order things correctly

No you don't! From the example in the Python docs:

```py
def parrot(voltage, state='a stiff', action='voom', type='Norwegian Blue'): ...

parrot(voltage=1000000, action='VOOOOOM')             # 2 keyword arguments
parrot(action='VOOOOOM', voltage=1000000)             # 2 keyword arguments
```

And even if it didn't, that doesn't stop us from allowing such behavior here.

---

Sorry, been a bit busy these past few days. I will try to review this tomorrow or on the weekend at the latest.

app/javascript/src/utils/compiler/parameterObjects.js
```diff
@@ -0,0 +1,70 @@
...
+  })
+
+  const result = { start, end, given }
+
+  // Return a false object to replace contents of unfound phrase
+  if (!completion) return {
+    ...result,
+    phraseParameters: [],
+    phraseDefaults: []
+  }
```
netux, line 52: Why would you want to replace the parameters if no completion is found?

Currently this would convert
```
NonExistant({ Foo: bar });
```

into
```
NonExistant();
```

Wouldn't it be less aggresive to just try and find the next parameter object? Even if the syntax is wrong for Workshop, I think the original code should still stay.

netux:
Doing the review now. One thing I noticed is parameter names are still highlighted like normal. I think they should be exempt (maybe by checking for a `:` after them?)


Note how "Button" in "Button:" is purpleish, whereas "Player" in "Player:" is the default color for variables.

app/javascript/src/utils/compiler/parameterObjects.js
```diff
@@ -0,0 +1,70 @@
...
+
+  return {
+    ...result,
+    phraseParameters: completion.detail_full.split(", "),
```
netux, line 56: Couldn't this completion object get a `parameter_names` property instead? Splitting the details string feels cursed.

spec/javascript/utils/compiler/parameterObjects.test.js
```diff
@@ -0,0 +1,150 @@
...
+    test("Should handle white space as expected", () => {
+      const expected = {
+        given: { One: "10", Three: "20" },
+        phraseParameters: ["One", "Two", "Three"],
+        phraseDefaults: [0, 0, 0]
+      }
+
+      expect(getFirstParameterObject("Some Action   ({ One: 10, Three: 20 })")).toEqual({ ...expected, start: 15, end: 36 })
+      expect(getFirstParameterObject("Some Action(   { One: 10, Three: 20 }   )")).toEqual({ ...expected, start: 15, end: 36 })
+      expect(getFirstParameterObject("Some Action({ One: 10,    Three: 20 })")).toEqual({ ...expected, start: 12, end: 36 })
```
netux, line 61: Also check for whitespace around the colons

```js
      expect(getFirstParameterObject("Some Action({ One    :10,    Three :     20 })")).toEqual({ ...expected, start: 12, end: 44 })
```

netux (approved):
LGTM. Code was surprisingly much shorter than I expected.

app/javascript/src/utils/compiler/parameterObjects.js
Mitcheljager: Yeah the problem was that the while loop finds the first result, so if you don't replace it it will find it again. I will try and find a solution, maybe I can start the loop from a later index or something

netux: Use [exec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec) instead of match

Mitcheljager: It was a bit trickier than that. The while loop is separate from the regex since the function is also used for one-offs like in the linting.
I opted to in the while loop check if the object was found, and if so set the start index from the index of that match. This index is given to the function and the string is sliced from there so it always starts after that match.
So if the first loop is `Fake({ a: b }) Real({ c: d })` than the second loop starts at `({ a: b }) Real({ c: d })`, and that offset is added to the start and end index of the second result.

netux (approved):

PR #393, opened 2023-12-22
Only apply ordered list counter style tolist items immediately after an unordered list

netux:
One-liner.

Given
```
1. ul li 1
  * ul li 1, ol li 1
  * ul li 2, ol li 2
2. ul li 2
3. ul li 3
  1. ul li 3, ul li 1
  2. ul li 3, ul li 2
4. ul li 4
  * ul li 4, ol li 1
```

Before
unordered list items increasing the counter of the parent ordered list

After
unordered list items not affecting the counter of the parent ordered list

### DIFF
--- a/app/assets/stylesheets/elements/_items.scss
+++ b/app/assets/stylesheets/elements/_items.scss
@@ -500,7 +500,7 @@
     list-style: none;
     padding-left: 2.5rem;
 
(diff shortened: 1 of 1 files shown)


Mitcheljager (approved):
Nice!

PR #395, opened 2024-01-08
Reimplement `findAllRules()` in Editor script importer

netux:
Use `RegExp.exec()` and `getClosingBracket()`. Now we create items for each rule, from where the rule starts being defined ("rule(...) {...") to the closing bracket.

While the old implementation was more lenient on what it included, this way we can fix a bug where the "disabled" tags on rules were being skipped over when importing.


Mitcheljager (approved):
Odd, but sure!

PR #399, opened 2024-01-22
Include subroutines used in actions in the compiled output of the Editor

netux:
While there is linter logic to catch these, it is sometimes hard to see the linter error in certain cases.

An argument can be made that, because the in-game editor allows for creating a subroutine but never defining a rule for it, we should also allow it as well.

In fact, this change fixes a case where importing a snippet that declares a subroutine but never define it, then compiling the project and trying to import that into the game would cause a workshop error because we stripped the subroutine from the subroutines list.

Also: refactor `getSubroutines` to use `String.matchAll()`, which is a bit better than `String.match()` and a ton of `String.replace()`s.


Mitcheljager (approved):
Nice, seems like a good fix

PR #400, opened 2024-01-28
 Add lint rule to the Editor that warns about usage of triple equals

netux:
Makes sure to only detect them outside of strings.

### DIFF
--- a/app/javascript/src/lib/OWLanguageLinter.js
+++ b/app/javascript/src/lib/OWLanguageLinter.js
@@ -633,7 +633,7 @@ function findTripleEquals(content) {
       from,
       to: from + match[0].length,
       severity: "error",
-      message: "This isn't JavaScript! Use single equals for assignments, or double equals to comparisons."
+      message: "Use single equals for assignments, or double equals to comparisons. Triple equals doesn't exist here (this isn't JavaScript!)."
     })
   }
 }
--- a/app/javascript/src/lib/OWLanguageLinter.js
+++ b/app/javascript/src/lib/OWLanguageLinter.js
@@ -1,4 +1,4 @@
-import { getClosingBracket, getPhraseFromPosition, splitArgumentsString } from "../utils/parse"
+import { findRangesOfStrings, getClosingBracket, getPhraseFromPosition, matchAllOutsideRanges, splitArgumentsString } from "../utils/parse"
 import { completionsMap, subroutinesMap, workshopConstants } from "../stores/editor"
 import { get } from "svelte/store"
 import { getFirstParameterObject } from "../utils/compiler/parameterObjects"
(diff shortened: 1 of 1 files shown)


app/javascript/src/lib/OWLanguageLinter.js
```diff
@@ -622,3 +623,17 @@ function findUndefinedSubroutines(content) {
     }
   }
 }
+
+function findTripleEquals(content) {
+  const stringRanges = findRangesOfStrings(content);
+
+  for (const match of matchAllOutsideRanges(stringRanges, content, /={3}/g)) {
```
Mitcheljager, line 630: While ={3} looks cool, `===` is much more readable

app/javascript/src/lib/OWLanguageLinter.js
```diff
@@ -622,3 +623,17 @@ function findUndefinedSubroutines(content) {
...
+    const from = match.index
+    diagnostics.push({
+      from,
+      to: from + match[0].length,
+      severity: "warning",
```
Mitcheljager, line 635: It should be an error rather than a warning since it results in an error in-game

app/javascript/src/lib/OWLanguageLinter.js
```diff
@@ -622,3 +623,17 @@ function findUndefinedSubroutines(content) {
...
+      message: "This isn't JavaScript! Use single equals for assignments, or double equals to comparisons."
```
Mitcheljager, line 636: While it's funny enough, for a lot of people this won't mean anything. I vote we leave out the "This isn't javascript!" bit

Mitcheljager (changes requested):
Good idea, but I do have a few suggestions

app/javascript/src/lib/OWLanguageLinter.js
netux:
```suggestion
      message: "Use single equals for assignments, or double equals to comparisons. Triple equals don't exist here (but they do in JavaScript)."
```

app/javascript/src/lib/OWLanguageLinter.js
netux:
```suggestion
  for (const match of matchAllOutsideRanges(stringRanges, content, /===/g)) {
```

app/javascript/src/lib/OWLanguageLinter.js
netux:
```suggestion
      severity: "error",
```

app/javascript/src/lib/OWLanguageLinter.js
Mitcheljager: Sure, that will do

Mitcheljager: Oh ha, the last one is better in my opinion

Mitcheljager (approved):

PR #404, opened 2024-02-07
fix: Include special characters in syntax highlighting

Mitcheljager:
Keywords with special accents were not being highlighted as expected. This PR fixes that by including these characters in the regex. It's a bit of a ChatGPT special, but it seems to work well.

Before | After
--- | ---
image | image


netux:
Did you give ChatGPT a list with the name of all heroes and asked it to give you the range of characters their names contain?

Everything in red is being captured by the new ranges. Do we really need Ø and ø?

Mitcheljager:
@netux I didn't, no. I just asked it to accept characters with accents. I suppose we don't really _need_ those characters, but they don't hurt either I suppose

netux:
> I suppose we don't really _need_ those characters, but they don't hurt either I suppose

In that case then why not add the rest of the latin alphabet? 😅

Mitcheljager:
@netux Because I didn't think of it! Updated it

netux (approved):

PR #406, opened 2024-02-07
Update comments.js

Loottoo:
Regex was eating the first character before '//' (usually a space or tab).

example: 
Global.cat = True;//cat
would compile without the ';'
Global.cat = True

Tested it here:
https://regex101.com/r/LvXuFG/1


Mitcheljager (approved):
Ah good find, I would have never noticed!

Loottoo:
> Ah good find, I would have never noticed!
neither would i, had i not ignored the comment while copy pasting code to test something 🙃

Mitcheljager:
@Loottoo It did break the tests though, would you like me to look at that or do you know what's up?

Loottoo:
Not sure whats up but i see the failed test, will have a look at it!

Loottoo:
@Mitcheljager The old regex removed the character before //, the new one wouldnt work without having a character before // (at start of line).
Asked chatGPT for a new one, hope it works better 😄 
but now the 🥒 failed instead..?

Mitcheljager:
If cucumber tests fail it's probably the tests being annoying, they fail randomly :/ I will rerun them

app/javascript/src/utils/compiler/comments.js
```diff
@@ -1,3 +1,3 @@
 export function removeComments(joinedItems) {
-  return joinedItems.replaceAll(/\/\*[\s\S]*?\*\/|([^:]|^)\/\/.*$/gm, "")
+  return joinedItems.replaceAll(/\/\*[\s\S]*?\*\/|(?:(?<=^)|(?<=[^:]))\/\/.*$/gm, "")
```
netux, line 2: Nice! That second positive lookbehind could be a negative lookbehind for `:` https://regex101.com/r/iZ4DRL/1
```suggestion
  return joinedItems.replaceAll(/\/\*[\s\S]*?\*\/|(?:(?<=^)|(?<!:))\/\/.*$/gm, "")
```

Or even better IMO, join both positive lookbehinds and remove the wrapping non-capturing group: https://regex101.com/r/rBo8D2/1
```suggestion
  return joinedItems.replaceAll(/\/\*[\s\S]*?\*\/|(?<=^|[^:])\/\/.*$/gm, "")
```

netux (approved):
LGTM. A few small suggestions, but they are not really necessary =)

app/javascript/src/utils/compiler/comments.js
Loottoo: Looks good to me 👍 i trust your regexing more than my own (well, chatGPT's) 😅

PR #407, opened 2024-02-08
fix(editor highlighter): match custom operators when they are not part of a word

netux:
Ensures operators are not surrounded by other letters.

Fixes cases like ```
Global.intersection
```
where the "in" in "intersection" would be matched like the custom operator `in`


Mitcheljager (approved):
👍

PR #409, opened 2024-02-08
Fixed several bugs with indentations

Loottoo:
* Empty lines containing only a linebreak are no longer removed upon shift-tabbing with several lines of code selected.
* Shift-tabbing after selecting a chunk of code at the start of a line no longer causes the selection to jump to the line above.
* Shift-tabbing after selecting a chunk of code at the very start of a file no longer results in "Uncaught RangeError: Invalid position -1".
* Shift-tabbing with the cursor in front of other code no longer selects the first character following the cursor.

Auto indent on linebreak:
* Pressing Enter while the cursor is located somewhere in the middle of indentation will no longer build more and more indents for each new line.
* Pressing Enter on a line with a open bracket now only indent if the cursor is inside the bracket.

### DIFF
--- a/app/javascript/src/components/editor/CodeMirror.svelte
+++ b/app/javascript/src/components/editor/CodeMirror.svelte
@@ -195,7 +195,9 @@
         if (shiftKey) {
           if (!/^\s/.test(insert[0]) && !(insert.includes("\n ") || insert.includes("\n\t"))) return { range: EditorSelection.range(from, to) }
 
-          insert = (/^\n/.test(insert[0]) ? "\n" : "").concat(insert.replaceAll(/\n[ \t]/g, "\n").substring(insert.search(/\S/) ? 1 : 0, insert.length))
+          const firstChar = insert[0]
+          insert = insert.replaceAll(/\n[ \t]/g, "\n").substring(insert.search(/\S/) ? 1 : 0, insert.length)
+          insert = (/^\n/.test(firstChar) ? "\n" : "").concat(insert)
         } else {
           insert = "\t" + insert.replaceAll("\n", "\n\t")
         }
--- a/app/javascript/src/components/editor/CodeMirror.svelte
+++ b/app/javascript/src/components/editor/CodeMirror.svelte
@@ -1,5 +1,5 @@
 <script>
-  import { onDestroy, onMount, createEventDispatcher } from "svelte"
+  import { onDestroy, onMount, createEventDispatcher, tick } from "svelte"
   import { basicSetup } from "codemirror"
   import { EditorView, keymap } from "@codemirror/view"
   import { EditorState, EditorSelection } from "@codemirror/state"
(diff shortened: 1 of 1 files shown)


app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -216,12 +217,13 @@
...
+  function getIndentForLine(state, line, charLimit) {
+    let lineText = state.doc.lineAt(Math.max(line, 0)).text
+    lineText = charLimit !== undefined ? lineText.slice(0, charLimit) : lineText
 
-    const spaces = /^\s*/.exec(state.doc.lineAt(line).text)?.[0].length
-    const tabs = /^\t*/.exec(state.doc.lineAt(line).text)?.[0].length
-    return Math.floor((spaces - tabs) / 4) + tabs
+    const tabs = /^\t*/.exec(lineText)?.[0].length
+    const spaces = /^\s*/.exec(lineText)?.[0].length - tabs
+    return Math.floor(spaces / 4) + tabs
```
Mitcheljager, line 226: I do feel like this was a little more readable before with the spaces subtracted in the return
```suggestion
    const tabs = /^\t*/.exec(lineText)?.[0].length
    const spaces = /^\s*/.exec(lineText)?.[0].length
    return Math.floor((spaces - tabs) / 4) + tabs
```

app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -195,12 +195,13 @@
         if (shiftKey) {
           if (!/^\s/.test(insert[0]) && !(insert.includes("\n ") || insert.includes("\n\t"))) return { range: EditorSelection.range(from, to) }
 
-          insert = insert.replaceAll(/\n[ \t]/g, "\n").substring(insert.search(/\S/) ? 1 : 0, insert.length)
+          insert = (/^\n/.test(insert[0]) ? "\n" : "").concat(insert.replaceAll(/\n[ \t]/g, "\n").substring(insert.search(/\S/) ? 1 : 0, insert.length))
```
Mitcheljager, line 198: This is a bit much to take in, maybe we can separate it to two lines

```suggestion
          insert = insert.replaceAll(/\n[ \t]/g, "\n").substring(insert.search(/\S/) ? 1 : 0, insert.length)
          insert = (/^\n/.test(insert[0]) ? "\n" : "").concat(insert)
```

Mitcheljager (changes requested):
Looks great overall, but I did have 2 suggestions

app/javascript/src/components/editor/CodeMirror.svelte
Loottoo: oh i changed it for the sake of making a bit more readable lol. i dont like how the 'spaces' constant is set to a wrong value just so you can make it correct by subtracting in the return. but the result is the same so you can use the previous if you want!

Mitcheljager: Oh true, tabs are included in the regex. Fair point!

app/javascript/src/components/editor/CodeMirror.svelte
Loottoo: yeah i agree, but not sure if thats going to work. if you make it two lines its going to `test(insert[0])` after insert has already had its linebreaks replaced right?

Mitcheljager: Hmm true, might need to write it to a new variable first, I'm not sure what the best solution would be

Loottoo: alright i made a new variable first. bit more code but its probably worth it 😃

Mitcheljager (approved):
Great!

PR #410, opened 2024-02-10
feat: Autocomplete using parameter objects

Mitcheljager:
This PR adds an options to the settings dropdown to autocomplete using parameter objects. All autocompletes will use parameter objects instead of their regular format. When you change the settings the completions map is regenerated with the new corresponding format.
Along with that settings there's a setting to only do this if the completion has more than x arguments, defaulting to 2. If it has less than the given amount the regular format will be used. This is mostly to prevent silly parameter objects like `Button({ Button: Melee })`.
param-objects-autocomplete

Since the autocompletions can be multi line I added a function that transforms transactions in CodeMirror that contain new lines to use the expected amount of tabs. An added benefit of this is that this fixes a bug with pasting code that contains tabs that would only be indented correctly on the first line

Before | After
--- | ---
paste-before | paste-after

### DIFF
--- a/app/javascript/src/components/editor/CodeMirror.svelte
+++ b/app/javascript/src/components/editor/CodeMirror.svelte
@@ -295,8 +295,7 @@
       const [range] = transaction.changedRanges
       const text = transaction.state.doc.toString().slice(range.fromB, range.toB)
       const indents = getIndentForLine(view.state, range.fromB)
-      let tabs = ""
-      for (let i = 0; i < indents; i++) { tabs += "\t" }
+      const tabs = "\t".repeat(indents)
 
       const changes = {
         from: range.fromB,
--- a/app/javascript/src/components/editor/CodeMirror.svelte
+++ b/app/javascript/src/components/editor/CodeMirror.svelte
@@ -2,9 +2,9 @@
   import { onDestroy, onMount, createEventDispatcher, tick } from "svelte"
   import { basicSetup } from "codemirror"
   import { EditorView, keymap } from "@codemirror/view"
-  import { EditorState, EditorSelection } from "@codemirror/state"
+  import { EditorState, EditorSelection, Transaction } from "@codemirror/state"
   import { indentUnit, StreamLanguage, syntaxHighlighting } from "@codemirror/language"
-  import { autocompletion } from "@codemirror/autocomplete"
+  import { autocompletion, pickedCompletion } from "@codemirror/autocomplete"
   import { redo } from "@codemirror/commands"
   import { linter, lintGutter } from "@codemirror/lint"
   import { indentationMarkers } from "@replit/codemirror-indentation-markers"
@@ -94,9 +94,12 @@
           { key: "Enter", run: autoIndentOnEnter },
           { key: "Ctrl-Shift-z", run: redoAction }
         ]),
-        EditorView.updateListener.of((state) => {
-          if (state.docChanged) updateItem()
-          if (state.selectionSet) $editorStates[currentId].selection = view.state.selection
+        EditorView.updateListener.of((transaction) => {
+          if (transaction.docChanged) {
+            indentMultilineInserts(transaction)
+            updateItem()
+          }
+          if (transaction.selectionSet) $editorStates[currentId].selection = view.state.selection
         }),
         basicSetup,
         parameterTooltip(),
@@ -225,6 +228,7 @@
 
     const tabs = /^\t*/.exec(lineText)?.[0].length
     const spaces = /^\s*/.exec(lineText)?.[0].length - tabs
+
     return Math.floor(spaces / 4) + tabs
   }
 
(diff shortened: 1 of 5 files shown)


app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -281,10 +285,28 @@
...
   }
 
+  function indentMultilineInserts(transaction) {
+    // Only perform this function if transaction is of an expected type performed by the user to prevent infinite loops on changes made by CodeMirror
+    if (transaction.transactions.every(tr => ["input.paste", "input.complete"].includes(tr.annotation(Transaction.userEvent)))) {
+      const [range] = transaction.changedRanges
+      const text = transaction.state.doc.toString().slice(range.fromB, range.toB)
+      const indents = getIndentForLine(view.state, range.fromB)
+      let tabs = ""
+      for (let i = 0; i < indents; i++) { tabs += "\t" }
```
netux, line 299:
```suggestion
      const tabs = "\t".repeat(indents)
```

app/javascript/src/stores/editor.js
```diff
@@ -104,5 +104,7 @@ export const settings = writable({
   "color-invalid": "#b33834",
   "color-custom-keyword": "#c678dd",
   "show-indent-markers": true,
-  "word-wrap": false
+  "word-wrap": false,
+  "autocomplete-parameter-objects": false,
+  "autocomplete-min-parameter-size": 2
```
netux, line 109: PR description says 3, but this is set to 2. Which one should it be?

app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -281,10 +285,28 @@
...
+  function setScrollPosition() {
     view.scrollDOM.scrollTo({ top: $editorScrollPositions[currentId] || 0 })
   }
 
+  function indentMultilineInserts(transaction) {
+    // Only perform this function if transaction is of an expected type performed by the user to prevent infinite loops on changes made by CodeMirror
+    if (transaction.transactions.every(tr => ["input.paste", "input.complete"].includes(tr.annotation(Transaction.userEvent)))) {
+      const [range] = transaction.changedRanges
+      const text = transaction.state.doc.toString().slice(range.fromB, range.toB)
+      const indents = getIndentForLine(view.state, range.fromB)
```
netux, line 297: Indentation is not all sunshine and rainbows, sadly

https://github.com/Mitcheljager/workshop.codes/assets/6181929/87e2c88f-6d4a-42b1-b462-196a1c28e0ff

I think this could benefit from reapplying indentation on each line of the pasted content?

app/javascript/src/components/editor/Editor.svelte
```diff
@@ -76,32 +77,37 @@
...
+      params.detail_full = joinedDetail
+      params.detail = `(${ joinedDetail.slice(0, 30) }${ joinedDetail.length > 30 ? "..." : "" })`
 
-          if (lowercaseDefaults.includes(string)) return defaults[toCapitalize(string)]
+      // Add apply values when selecting autocomplete, filling in default args
+      const lowercaseDefaults = Object.keys(defaults).map(k => k.toLowerCase())
+      const useParameterObject = $settings["autocomplete-parameter-objects"] && params.args_length >= $settings["autocomplete-min-parameter-size"]
+      const apply = v.args.map(a => {
+        let string = a.default?.toString().toLowerCase().replaceAll(",", "")
+        if (useParameterObject) string = `\n\t${ a.name }: ${ string }`
```
netux, line 94: Would it make sense to have a limit on when each parameter is put on a new line and when not?

If I was one of those people that put `autocomplete-min-parameter-size` to 1, maybe I'd like for autocomplete to not expand the parameter object into multiple lines if the parameter amount is <3 or something.

netux (commented):
Pretty good! The indentation changes don't work so well, but the autocomplete is greatly appreciated 🙂

app/javascript/src/stores/editor.js
Mitcheljager: Ah yes, it should be 2. I had it at 3 initially, but 2 felt better

app/javascript/src/components/editor/CodeMirror.svelte
Mitcheljager: Ah, neat!

app/javascript/src/components/editor/CodeMirror.svelte
Mitcheljager: I made some improvements and it should be much better. It's not quite perfect but the imperfections make enough sense. VSCode has similar (but different) inconsistencies. I think this is about as close as we can get without fully parsing the text, which would be too slow for larger texts anyway.

app/javascript/src/components/editor/Editor.svelte
Mitcheljager: Good idea!

min-newline

PR #411, opened 2024-02-11
refactor: Move indent function to separate file

Mitcheljager:
## This merges into #410 

`CodeMirror.svelte` was becoming too large and hard to process. This PR moves all indent related function to a separate file. Functions are largely unchanged but I had to make some small changes to accommodate the move. I've also included JSDocs for all of the functions. Additionally I've moved some functions in `CodeMirror.svelte` around to make a little more sense.

### DIFF
--- a/app/javascript/src/components/editor/CodeMirror.svelte
+++ b/app/javascript/src/components/editor/CodeMirror.svelte
@@ -122,7 +122,7 @@
   }
 
   /**
-   * Returns a plguin that updates a store of scroll positions when the view is scrolled.
+   * Returns a plugin that updates a store of scroll positions when the view is scrolled.
    * The view is also scrolled when updating the view, but we don't want to store that position.
    * For this we use the updatingState flag to determine if it was a user scroll or a update scroll.
    */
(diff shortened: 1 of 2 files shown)


app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -69,48 +117,32 @@
...
-        }),
-        basicSetup,
-        parameterTooltip(),
-        indentationMarkers(),
-        rememberScrollPosition(),
-        foldBrackets(),
-        ...($settings["word-wrap"] ? [EditorView.lineWrapping] : [])
-      ]
+  /**
+   * Returns a plguin that updates a store of scroll positions when the view is scrolled.
```
netux, line 125: Typo

```suggestion
   * Returns a plugin that updates a store of scroll positions when the view is scrolled.
```

app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -69,48 +117,32 @@
 
     requestAnimationFrame(() => {
       updatingState = false
-      setScrollPosition()
+      setScrollPosition(view, currentId)
     })
   }
```
netux, line 122: Was this really the best way to do this? Why do we need to overwrite the scroll position each frame, and not only when the editor is being initialized/the user goes to another file?

netux (approved):
Sure! I'm wondering if that `rememberScrollPosition` plugin could be extracted to its own file too. Seems like its dependencies (`$editorScrollPositions` and `updatingState`) are needed only for it as well.

app/javascript/src/components/editor/CodeMirror.svelte
Mitcheljager: True, it might be a bit much. Since it's not related to this PR I will ignore for now, but I will create an issue so we don't forget

PR #412, opened 2024-02-12
Avoid checking for incorrect argument count when the function used is a Workshop constants in the Editor linter

netux:


Mitcheljager (approved):
That will do!

PR #414, opened 2024-02-12
Remove [linemarker]s from parameter object keys in the Editor compiler

### LINKED ISSUE #402: [Bug] [Editor] Compiled code in line number search dialog always replaces parameter objects with the default parameter values
In the image above, I have a `Small Message` using a parameter object to set `Visible To` to `Empty Array` and `Header` to `Custom String("Non-default text")`. However, in the compiled output preview, the parameter object is replaced with the values of `All Players(All Teams)` and `Custom String("Some string")` respectively, which are the default values for these parameters.

netux:
Fixes an issue where parameter keys would internally contain the [linemarker]s, meaning we wouldn't find the correct argument positions and just use default values instead.

Fixes #402


Mitcheljager (approved):
That will do it!

PR #416, opened 2024-02-13
missed a 0 in last fix

Loottoo:
when did the zero fall away and how did i not notice it missing? 🤔
reapplying the 0 so that it actually does something


Mitcheljager (approved):
Oh lol whoops :D

Loottoo:
Yeah... 😅

PR #417, opened 2024-02-16
Adjustments for auto-indents when pasting code

Loottoo:
Further adjustments to the code for auto-indents! the goal is to allow pasting of whitespace characters, to make copy-pasting on the same line work as it used to (duplicate line), and to keep the structure of indents for code that is either pasted in a marked area or on lines that already have code. the behavior should be fairly close to what is used in vscode.

### DIFF
--- a/app/javascript/src/utils/codemirror/indent.js
+++ b/app/javascript/src/utils/codemirror/indent.js
@@ -1,5 +1,4 @@
 import { EditorSelection, Transaction } from "@codemirror/state"
-import {indent} from "@codemirror/state"
 /**
  * Indent on using tab with special conditions. Indent is reversed while holding shift
  * @param {Object} view CodeMirror view
--- a/app/javascript/src/utils/codemirror/indent.js
+++ b/app/javascript/src/utils/codemirror/indent.js
@@ -1,5 +1,5 @@
 import { EditorSelection, Transaction } from "@codemirror/state"
-
+import {indent} from "@codemirror/state"
 /**
  * Indent on using tab with special conditions. Indent is reversed while holding shift
  * @param {Object} view CodeMirror view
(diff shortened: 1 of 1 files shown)


app/javascript/src/utils/codemirror/indent.js
```diff
@@ -141,31 +140,28 @@ export function indentMultilineInserts({ state, dispatch }, transaction) {
...
-    let firstIndentCount = 0
-    const mappedText = splitText.map((line, i) => {
-      if (!i) {
-        firstIndentCount = getIndentCountForText(line)
-        startIndentCount = getIndentCountForText(rangeLine.text) - firstIndentCount
-
-        return line.replace(/^\s+/, "")
+    let indentDifference = 0
+    //Dont touch single-line content (allow pasting of whitespace characters without treating them as indents)
+    if(splitText.length >= 2){
```
Mitcheljager, line 148: I think ultimately while this helps, it still points towards larger issues. When I wrote this it was originally only meant for autocomplete, and fixing some pasting business was just a nice bonus, but it's proven to be less than helpful. Can I propose we just disable it for pasting and only keep it for autocomplete?

Loottoo: yep i see your point! its pretty clear that the pasting is janky because of inconsistencies elsewhere. and im not a fan of whitespace characters being treated as indents (and possibly getting removed). 
might be best to use the code as it currently is live for the autocomplete, and then make only a minor adjustment to the paste without changing what is actually getting pasted. i still want to deal with those wrong indents on the first line though. just leave the PR laying around for now and i will have a look at it tomorrow 😀

Mitcheljager: CodeMirror already did a decent job of automatically indenting pastes, my attempts to make it better made it worse :D

Removing it should be as simple as removing `"input.paste"` from https://github.com/Mitcheljager/workshop.codes/pull/417/files#diff-57562ceac603867d58c8cbfb32b690dde45b10366da68ac14eb56f19b422b8a5R141

Loottoo: Alright lets just take the most straightforward and simplest solution then! But i still want to have a look at that first indent at one point or another :P

Mitcheljager: Sure!

PR #418, opened 2024-02-18
Prevent looping when no closing parenthesis is found

Loottoo:
linterOops
The argumentsString variable builds up to extreme lengths when argumentClosing is -1 (no closing parenthesis).

Reproduce with code:

```hs
rule("Each Player Rule")
{
    event
    {
        Ongoing - Each Player;
        All;
        All;
    }

    conditions
    {

    }

    actions
    {
	    Apply Impulse(Event Player, Dot Product(Velocity Of(
	    Event Player), Normalize(X Component Of(Throttle))

        
    }
}
```

### DIFF
--- a/app/javascript/src/lib/OWLanguageLinter.js
+++ b/app/javascript/src/lib/OWLanguageLinter.js
@@ -122,7 +122,7 @@ function findIncorrectArgsLength(content) {
         let argumentMatch
         while ((argumentMatch = /\(/g.exec(argumentsString)) != null && safeIndex < 100) {
           const argumentClosing = getClosingBracket(argumentsString, "(", ")", argumentMatch.index - 1)
-          if (argumentClosing === -1) { break }
+          if (argumentClosing === -1) break
           argumentsString = argumentsString.substring(0, argumentMatch.index) + argumentsString.substring(argumentClosing + 1)
 
           safeIndex++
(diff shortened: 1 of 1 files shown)


app/javascript/src/lib/OWLanguageLinter.js
```diff
@@ -122,6 +122,7 @@ function findIncorrectArgsLength(content) {
         let argumentMatch
         while ((argumentMatch = /\(/g.exec(argumentsString)) != null && safeIndex < 100) {
           const argumentClosing = getClosingBracket(argumentsString, "(", ")", argumentMatch.index - 1)
+          if (argumentClosing === -1) { break }
```
netux, line 125: I feel like Mitsiee would ask for this, so let me ask for it first 😁

```suggestion
          if (argumentClosing === -1) break
```

Loottoo: are there coding standards for this that should be followed or is it just personal preference? 
i assume its because its single line that the clamps shouldnt be there? 
if nothing else i suppose its better to remove them just for the sake of keeping it consistent with the rest of the code

netux: I guess its Mitsiee's personal style. That is to say, its the code style of the project since he's the owner 😄.

And indeed this is because the if-statement only contains one statement inside. I thought ESLint would catch this, but I guess not.

Mitcheljager: Indeed, it's just a preference things. I will see if I can set up ESLint or prettier a little better, ideally they would catch all of these things

Mitcheljager (approved):

Issue #246, opened 2023-02-04
[Editor] Autocomplete does not insert a semicolon


Mitcheljager:
When autocompleting something no semicolon is inserted, requiring to type it yourself. In cases where the results go offscreen (for example using `Create HUD Text` it can be easy to miss. The editor will tell you the semicolon is missing, but it's still easy to ignore. It would be nice if it inserted it automatically.

One difficulty here is that it doesn't need to insert semicolons if the when autocompleting something nested. One solution could be to check if it's inside of brackets (`[,(,),],{,}`), or perhaps doing it just for actions and not for values might be good enough?

PR #419, opened 2024-02-20
Fix default for parameter objects & semicolon after autocomplete

Loottoo:
The default values inserted when using autocomplete and parameter objects were not correct. for example, when you expect to get the number 0 you would instead get the string "Number". i changed the code to handle the name and default separately before adding the newlines which did the trick.

Also added semicolons to autocomplete according to issue [#246](https://github.com/Mitcheljager/workshop.codes/issues/246).
The semicolons are inserted only when autocompleting actions. having it insert a semicolon for other type of values could probably get annoying

### DIFF
--- a/app/javascript/src/components/editor/CodeMirror.svelte
+++ b/app/javascript/src/components/editor/CodeMirror.svelte
@@ -17,6 +17,7 @@
   import { translationsMap } from "../../stores/translationKeys"
   import { getPhraseFromPosition } from "../../utils/parse"
   import { tabIndent, getIndentForLine, getIndentCountForText, shouldNextLineBeIndent, autoIndentOnEnter, indentMultilineInserts } from "../../utils/codemirror/indent"
+  import { get } from "svelte/store"
   import debounce from "../../debounce"
 
   const dispatch = createEventDispatcher()
@@ -70,7 +71,7 @@
         ]),
         EditorView.updateListener.of((transaction) => {
           if (transaction.docChanged) {
-            indentMultilineInserts(view, transaction)
+            autocompleteFormatting(view, transaction)
             updateItem()
           }
           if (transaction.selectionSet) $editorStates[currentId].selection = view.state.selection
@@ -162,6 +163,34 @@
       to: word.to,
       options: specialOverwrite || [...$completionsMap, ...$variablesMap, ...$subroutinesMap, ...extraCompletions],
       validFor: /^(?:[a-zA-Z0-9]+)$/i
+    }
+  }
+
+  function autocompleteFormatting(view, transaction) {
+    if (transaction.transactions.every(tr => ["input.complete"].includes(tr.annotation(Transaction.userEvent)))) {
+      indentMultilineInserts(view, transaction)
+      insertSemicolon(view, transaction)
+    }
+  }
+
+  function insertSemicolon(view, transaction) {
+    const line = view.state.doc.lineAt(transaction.changedRanges[0].fromB)
+    const phrase = getPhraseFromPosition(line, transaction.changedRanges[0].fromB)
+
+    const possibleValues = get(completionsMap).filter(v => v.args_length)
+    const validValue = possibleValues.find(v => v.label == phrase.text)
+
+    if (validValue?.type) {
+      const insertPosition = view.state.selection.ranges[0].from
+      
+      if(validValue.type === "function") {
+        view.dispatch({
+          changes: { from: insertPosition, insert: `;` },
+          selection: EditorSelection.create([
+            EditorSelection.range(insertPosition + 1, insertPosition + 1)
+          ])
+        })
+      }
     }
   }
 
--- a/app/javascript/src/components/editor/CodeMirror.svelte
+++ b/app/javascript/src/components/editor/CodeMirror.svelte
@@ -174,17 +174,17 @@
   }
 
   function insertSemicolon(view, transaction) {
+    if (!$settings["autocomplete-semicolon"]) return
+
     const line = view.state.doc.lineAt(transaction.changedRanges[0].fromB)
     const phrase = getPhraseFromPosition(line, transaction.changedRanges[0].fromB)
-
     const possibleValues = get(completionsMap).filter(v => v.args_length)
     const validValue = possibleValues.find(v => v.label == phrase.text)
 
     if (!validValue?.type) return
+    if(validValue.type !== "function") return
 
     const insertPosition = view.state.selection.ranges[0].from
-      
-    if(validValue.type !== "function") return
     
     view.dispatch({
       changes: { from: insertPosition, insert: ";" },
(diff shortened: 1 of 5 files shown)


app/javascript/src/components/editor/Editor.svelte
```diff
@@ -92,12 +92,13 @@
       const useNewlines = params.args_length >= $settings["autocomplete-min-parameter-newlines"]
 
       const apply = v.args.map(a => {
-        let string = a.default?.toString().toLowerCase().replaceAll(",", "")
-        if (useParameterObject) string = `${ useNewlines ? "\n\t" : "" }${ a.name }: ${ string }`
+        const name = toCapitalize(a.name?.toString().toLowerCase())
+        let defaultValue = a.default?.toString().toLowerCase().replaceAll(",", "")
+        defaultValue = lowercaseDefaults.includes(defaultValue.toLowerCase()) ?
+          defaults[toCapitalize(defaultValue)] :
+          toCapitalize(defaultValue)
```
Mitcheljager, line 99: This is a bit hard to read, what do you thing of something like this?

```suggestion
        let defaultValue = a.default?.toString().toLowerCase().replaceAll(",", "")
        
        if (lowercaseDefaults.includes(defaultValue.toLowerCase()) defaultValue = defaults[toCapitalize(defaultValue)]
        else defaultValue =toCapitalize(defaultValue)
```

app/javascript/src/components/editor/Editor.svelte
```diff
@@ -92,12 +92,13 @@
...
 
-        if (lowercaseDefaults.includes(string)) return defaults[toCapitalize(string)]
-
-        return toCapitalize(string)
+        return useParameterObject ? `${ useNewlines ? "\n\t" : "" } ${ name }: ${ defaultValue }` : defaultValue
```
Mitcheljager, line 101: In a similar vain, what do you think of doing this instead?

```suggestion
        if (useParameterObject) return `${ useNewlines ? "\n\t" : "" } ${ name }: ${ defaultValue }` 
        
        return defaultValue
```

app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -165,6 +166,34 @@
...
+      
+      if(validValue.type === "function") {
+        view.dispatch({
+          changes: { from: insertPosition, insert: ";" },
+          selection: EditorSelection.create([
+            EditorSelection.range(insertPosition + 1, insertPosition + 1)
+          ])
+        })
+      }
+    }
```
Mitcheljager, line 194: Perhaps we can make both of these early returns to make it a bit less pyramid-y

```suggestion
    if (!validValue?.type) return

    const insertPosition = view.state.selection.ranges[0].from
      
    if(validValue.type !== "function") return
    
    view.dispatch({
      changes: { from: insertPosition, insert: ";" },
      selection: EditorSelection.create([
        EditorSelection.range(insertPosition + 1, insertPosition + 1)
      ])
    })
```

Mitcheljager (changes requested):
Looks great! I did have a couple of formatting suggestions

app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -165,6 +166,34 @@
     }
   }
 
+  function autocompleteFormatting(view, transaction) {
+    if (transaction.transactions.every(tr => ["input.complete"].includes(tr.annotation(Transaction.userEvent)))) {
+      indentMultilineInserts(view, transaction)
+      insertSemicolon(view, transaction)
```
netux, line 172: Could this maybe be optional? 😁

netux (approved):
LGTM

app/javascript/src/components/editor/Editor.svelte
Loottoo: i feel like if the first return is within the if then the second return should be in the `else`?

app/javascript/src/components/editor/CodeMirror.svelte
netux: Just don't place the semicolon in any situation. Make it so I have to add it myself (because I may have the reflex of doing so automatically).

I agree on that such a setting may clutter the settings section. But maybe its time to make that into a separate dialog (a.k.a. on another PR :D)

app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -174,17 +174,17 @@
   }
 
   function insertSemicolon(view, transaction) {
+    if (!$settings["autocomplete-semicolon"]) return
```
netux, line 177: Wouldn't it make more sense to just not call the function?

```ts
  function autocompleteFormatting(view, transaction) {
    if (transaction.transactions.every(tr => ["input.complete"].includes(tr.annotation(Transaction.userEvent)))) {
      indentMultilineInserts(view, transaction)
      if ($settings["autocomplete-semicolon"])
          insertSemicolon(view, transaction)
    }
  }
```

Loottoo: yes, not sure why it put it the function

app/javascript/src/components/editor/Settings.svelte
```diff
@@ -81,6 +81,15 @@
         </div>
       </div>
 
+      <div class="checkbox tooltip mt-1/8">
+        <input id="autocomplete-semicolon" type="checkbox" bind:checked={$settings["autocomplete-semicolon"]} />
```
Mitcheljager, line 85: This setting also needs to be added to the defaults further up the file, it should default to true in my opinion

app/javascript/src/components/editor/Editor.svelte
```diff
@@ -92,12 +92,15 @@
...
-        if (useParameterObject) string = `${ useNewlines ? "\n\t" : "" }${ a.name }: ${ string }`
-
-        if (lowercaseDefaults.includes(string)) return defaults[toCapitalize(string)]
-
-        return toCapitalize(string)
+        const name = toCapitalize(a.name?.toString().toLowerCase())
+        let defaultValue = a.default?.toString().toLowerCase().replaceAll(",", "")
+        
+        if (lowercaseDefaults.includes(defaultValue.toLowerCase())) defaultValue = defaults[toCapitalize(defaultValue)]
+        else defaultValue =toCapitalize(defaultValue)
```
Mitcheljager, line 99: This was my fault in a suggestion 😬 
```suggestion
        else defaultValue = toCapitalize(defaultValue)
```

app/javascript/src/components/editor/Settings.svelte
netux: It was, wasn't it? https://github.com/Mitcheljager/workshop.codes/pull/419/commits/6edec229360f7de2d6799234d2800618330cc156#diff-a3c60ef0735190d4f10ce80e4f0fa56665aebc6d548cd541449b0e3e4bf4c347R108

Mitcheljager: Ah yes, my bad

Mitcheljager (approved):

netux (approved):

PR #421, opened 2024-02-21
Update actions and values list in Editor grammar files

netux:
Fixes `Start Healing Modification` not being highlighted, and probably some other things too.

Also adds instructions on how to generate these lists.


Mitcheljager (approved):
Sure! The grammar file isn't actually used, but that's ok

PR #422, opened 2024-02-21
Update default values for parameters to match the default values from the in-game editor

netux:
- `Count Of` → `Count Of(Array())` (added parenthesis)
- `String` → `Custom String("")` (removed "Some string" text)

This should work nicely alongside #419 :)


netux:
Another choice for `Count Of` would be `Count Of(Global.variable)` (as the in-game editor defaults to `Count Of(Global.A)`) for `For Global Variable`/`For Player Variable`. But either choice takes about the same amount of Backspace presses to remove (assuming you start at the end and hold Ctrl), so 🤷.

Mitcheljager (approved):
Nice, much better

PR #423, opened 2024-02-23
Exclude vector from parameter object autocompleting

Loottoo:
i dont like how vectors autocomplete as parameter objects when using less than 4 Minimum parameter length.

Vector({
X: 0,
Y: 0,
Z: 0
})

Vectors within other objects already autocomplete in a single line: 
Distance Between({
Start Pos: Vector(0, 0, 0),
End Pos: Vector(0, 0, 0)
})

As for the actual coding, not sure if this is the best way to do it? checking for x, y, and z individually doesnt make for the prettiest code. could also check if (.lenght === 1) as the X, Y, Z are the only names with a single letter. i doubt they will add more values with single letter names, but maybe?

logging some values to console:

`if(name.length === 1) console.log("name: " + name)`
name: X
name: Y
name: Z

`if(detail[0].length === 1) console.log("detail: " + detail)`
detail: X,Y,Z


Mitcheljager:
Hmm, I get your point but I'm not sure exceptions like this are the best way to go

Loottoo:
realized i can just check if its a vector earlier and not check for x, y, z :p
anyway, its fine if you want to keep it as is. i will leave it up to you 👍

Mitcheljager:
@Loottoo It's mostly inconsistent in how it would feel to the user. I agree I wouldn't want vectors to use parameter objects, but some people might. Without any context as to why it doesn't happen for Vectors it would feel more like a bug than an intended feature. I think if we want to do this we need to go a little further and the users select which actions they want to exclude from parameter objects.

Loottoo:
you make a fair point as always. lets take some time and consider the alternatives then. 
should i just close the PR?

Mitcheljager:
@Loottoo Could you convert it to a draft instead? Just so it's still there and we won't forget :D

PR #424, opened 2024-02-25
evaluate parameter objects in getVariables

Loottoo:
variables were not being added correctly when using parameter objects, sometimes causing errors when variables end up being added as undefined.

repro: autocomplete `For Player Variable` or `Chase Player Variable At Rate` as parameter objects, then try autocompleting anything.

### DIFF
--- a/app/javascript/src/utils/compiler/variables.js
+++ b/app/javascript/src/utils/compiler/variables.js
@@ -1,5 +1,5 @@
 import { findRangesOfStrings, getClosingBracket, matchAllOutsideRanges, splitArgumentsString } from "../parse"
-
+import { evaluateParameterObjects } from "./parameterObjects"
 // NOTE: The fact variable names can start with a decimal is intention.
 // We leave it to Overwatch to warn the user that this is not allowed.
 const globalVariablesRegex = /(?<=Global\.)[A-Za-z0-9_]+/g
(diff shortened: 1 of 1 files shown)


netux (approved):
LGTM!

app/javascript/src/utils/compiler/variables.js
```diff
@@ -1,5 +1,5 @@
 import { findRangesOfStrings, getClosingBracket, matchAllOutsideRanges, splitArgumentsString } from "../parse"
-
+import { evaluateParameterObjects } from "./parameterObjects"
 // NOTE: The fact variable names can start with a decimal is intention.
```
netux, line 3: For formatting purposes

```suggestion
import { evaluateParameterObjects } from "./parameterObjects"

// NOTE: The fact variable names can start with a decimal is intention.
```

app/javascript/src/utils/compiler/variables.js
```diff
@@ -138,6 +138,7 @@ function getLiteralPlayerVariables(source, stringRanges) {
 }
 
 export function getVariables(joinedItems) {
+  joinedItems = evaluateParameterObjects(joinedItems)
```
netux, line 141: I think we should add checks for this in variables.test.js too

netux: Yep, and you can run them locally with
```bash
npx jest
```

Or there are also a few VSCode extensions that add UI to run them, like [firsttris.vscode-jest-runner](https://marketplace.visualstudio.com/items?itemName=firsttris.vscode-jest-runner)

PR #427, opened 2024-03-01
Indent on paste and search adjustments

Loottoo:
**pasteIndentAdjustments** Should be pretty friendly. only changes actual tabs and spaces so specialcharacters are not removed, and only adjusts the first line since codemirror does a decent job with the rest. (The previous indent changes were not applied to paste since the code was changed to adjust indents for autocomplete separately).

**searchScrollMargin** 
Makes sure the searched results are not hidden behind the search window.

**_matches.scss** and **FindReplaceAll.svelte**:
Make the items that are excluded from compiler show with line-through text in the search.

**_codemirror.scss**:
Show which of the search results are actually selected.

**_editor.scss**:
Make the scrollbar easier to click.

**_editor-item.scss**:
Make it easier to click the beginning of the item name, and prevent icon tooltips from going underneath the selected item.

Feel free to criticize if something is not readable enough or if something needs adjustments (especially in Codemirror.svelte) 😬


app/assets/stylesheets/editor/_matches.scss
```diff
@@ -30,3 +30,8 @@
     color: $white;
   }
 }
+
+.hidden__item {
```
Mitcheljager, line 34: This isn't quite how BEM works. `[b]lock__[e]lement--[m]odifier`, so in this case there is no `hidden` block. A better name might be `matches__hidden` or `matches__item--hidden` if the class is applied directly to the same element rather than in a separate span.

app/assets/stylesheets/elements/_codemirror.scss
```diff
@@ -113,6 +113,11 @@
   .cm-searchMatch {
     background: lighten($bg-dark, 10%);
     box-shadow: 0 0 0 1px $primary;
+	
+    &-selected {
+      background: darken($bg-dark, 10%);
+      box-shadow: 0 0 0 1px $shadow;
+    }
   }
```
Mitcheljager, line 121: While classes like this work, they make it hard to search for. With BEM it's easier since you know the block level, but codemirror does not use bem.

```suggestion
  .cm-searchMatch {
    background: lighten($bg-dark, 10%);
    box-shadow: 0 0 0 1px $primary;
  }
	
  .cm-searchMatchselected {
    background: darken($bg-dark, 10%);
    box-shadow: 0 0 0 1px $shadow;
  }
```

app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -86,6 +87,19 @@
     })
   }
 
+  function handleTransactionDocChanged(view, transaction) {
+    if (transaction.docChanged) {
+      // Only perform this function if transaction is of an expected type performed by the user
+      // to prevent infinite loops on changes made by CodeMirror.
+      const userEvents = transaction.transactions.map(tr => tr.annotation(Transaction.userEvent))
+      if (userEvents.every(eventType => ["input.complete"].includes(eventType))) {
```
Mitcheljager, line 95: I was a bit lazy previously, but this notation can now just be 
```suggestion
      if (userEvents.every(eventType => eventType === "input.complete")) {
```

app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -86,6 +87,19 @@
...
+        autocompleteFormatting(view, transaction)
+      } else if (userEvents.every(eventType => ["input.paste"].includes(eventType))) {
```
Mitcheljager, line 97: And the same here

```suggestion
      } else if (userEvents.every(eventType => eventType === "input.paste")) {
```

app/javascript/src/components/editor/CodeMirror.svelte
Mitcheljager, line 95:
```suggestion
      const userEvents = transaction.transactions.map(tr => tr.annotation(Transaction.userEvent))
      
      if (userEvents.every(eventType => ["input.complete"].includes(eventType))) {
```

app/javascript/src/components/editor/CodeMirror.svelte
```diff
@@ -144,6 +158,16 @@
     view.scrollDOM.scrollTo({ top: $editorScrollPositions[id] || 0 })
   }
 
+  function searchScrollMargin(view, transaction) {
+    if (transaction.transactions[0].annotations[0].value !== "select.search") return
+    view.dispatch ({
```
Mitcheljager, line 163:
```suggestion
    if (transaction.transactions[0].annotations[0].value !== "select.search") return
    
    view.dispatch ({
```

app/javascript/src/utils/codemirror/indent.js
```diff
@@ -167,3 +167,27 @@ export function indentMultilineInserts({ state, dispatch }, transaction) {
...
+ * @param {Object} transaction CodeMirror transaction
+ */
+export function pasteIndentAdjustments({ dispatch }, transaction){
+  const [range] = transaction.changedRanges
+  const paste = transaction.state.doc.toString().slice(range.fromB, range.toB)
+  const line = transaction.state.doc.lineAt(range.fromB)
+  const lineText = line.text.slice(0, range.fromB - line.from)
+
+  let indentCount = 0
+  if(paste.includes("\n")){
```
Mitcheljager, line 183:
```suggestion
  if (paste.includes("\n")) {
```

app/javascript/src/utils/codemirror/indent.js
```diff
@@ -167,3 +167,27 @@ export function indentMultilineInserts({ state, dispatch }, transaction) {
...
+    while((lineText[indentCount] == "\t" || lineText[indentCount] == " ") &&
```
Mitcheljager, line 184:
```suggestion
    while ((lineText[indentCount] == "\t" || lineText[indentCount] == " ") &&
```

app/javascript/src/utils/codemirror/indent.js
```diff
@@ -167,3 +167,27 @@ export function indentMultilineInserts({ state, dispatch }, transaction) {
...
+    (paste[indentCount] == "\t" || paste[indentCount] == " ")){
```
Mitcheljager, line 185:
```suggestion
    (paste[indentCount] == "\t" || paste[indentCount] == " ")) {
```

app/assets/stylesheets/editor/_editor.scss
```diff
@@ -136,7 +136,7 @@ $top-height: 4rem;
   @include media-min(mbl) {
     cursor: ew-resize;
     height: 100%;
-    width: 0.5rem;
+    width: 0.25rem;
```
Mitcheljager, line 139: This does make it really quite small, but perhaps that's innevitable

Mitcheljager (changes requested):
Looks great! I primarily have some formatting suggestions

PR #428, opened 2024-03-02
Indent on paste and search adjustments

Loottoo:
**pasteIndentAdjustments** Should be pretty friendly. only changes actual tabs and spaces so specialcharacters are not removed, and only adjusts the first line since codemirror does a decent job with the rest. (The previous indent changes were not applied to paste since the code was changed to adjust indents for autocomplete separately).

**searchScrollMargin**
Makes sure the searched results are not hidden behind the search window.

**_matches.scss** and **FindReplaceAll.svelte**:
Make the items that are excluded from compiler show with line-through text in the search.

**_codemirror.scss**:
Show which of the search results are actually selected.

**_editor.scss**:
Make the scrollbar easier to click.

**_editor-item.scss**:
Make it easier to click the beginning of the item name, and prevent icon tooltips from going underneath the selected item.

Feel free to criticize if something is not readable enough or if something needs adjustments (especially in Codemirror.svelte) 😬

### DIFF
--- a/app/assets/stylesheets/editor/_editor-item.scss
+++ b/app/assets/stylesheets/editor/_editor-item.scss
@@ -23,7 +23,7 @@
     top: 0.25rem;
     left: 0.45rem;
     height: 1.5rem;
-    width: 1.5rem;
+    width: 1.05rem;
     color: darken($gray, 30%);
   }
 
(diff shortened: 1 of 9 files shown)


Mitcheljager (approved):
Looks good!

PR #429, opened 2024-03-04
evaluate parameter objects in getVariables and tests

Loottoo:
variables were not being added correctly when using parameter objects, sometimes causing errors when variables end up being added as undefined.

repro: autocomplete For Player Variable or Chase Player Variable At Rate as parameter objects, then try autocompleting anything.

### DIFF
--- a/app/javascript/src/utils/compiler/variables.js
+++ b/app/javascript/src/utils/compiler/variables.js
@@ -1,5 +1,5 @@
 import { findRangesOfStrings, getClosingBracket, matchAllOutsideRanges, splitArgumentsString } from "../parse"
-
+import { evaluateParameterObjects } from "./parameterObjects"
 // NOTE: The fact variable names can start with a decimal is intention.
 // We leave it to Overwatch to warn the user that this is not allowed.
 const globalVariablesRegex = /(?<=Global\.)[A-Za-z0-9_]+/g
@@ -138,6 +138,7 @@ function getLiteralPlayerVariables(source, stringRanges) {
 }
 
 export function getVariables(joinedItems) {
+  joinedItems = evaluateParameterObjects(joinedItems)
   const stringRanges = findRangesOfStrings(joinedItems)
 
   const playerVariablesFromActions = []
--- a/app/javascript/src/utils/compiler/variables.js
+++ b/app/javascript/src/utils/compiler/variables.js
@@ -1,6 +1,7 @@
 import { findRangesOfStrings, getClosingBracket, matchAllOutsideRanges, splitArgumentsString } from "../parse"
 import { evaluateParameterObjects } from "./parameterObjects"
-// NOTE: The fact variable names can start with a decimal is intention.
+
+// NOTE: The fact variable names can start with a decimal is intentional.
 // We leave it to Overwatch to warn the user that this is not allowed.
 const globalVariablesRegex = /(?<=Global\.)[A-Za-z0-9_]+/g
 
@@ -13,7 +14,6 @@ const globalVariablesRegex = /(?<=Global\.)[A-Za-z0-9_]+/g
  *   "Global.myGlobalVar.myPlayerVar.otherPlayerVar" => ".myPlayerVar", ".otherPlayerVar"
  *   "Global.Global.Global.myPlayerVar.otherPlayerVar" => "Global" (the second one), ".myPlayerVar", ".otherPlayerVar"
  *
- * NOTE: The fact variable names can start with a
  */
 const possiblePlayerVariablesRegex = /(?<![^\w.]Global)\.(?<variableName>[A-Za-z0-9_]+)/g
 
--- a/spec/javascript/utils/compiler/variables.test.js
+++ b/spec/javascript/utils/compiler/variables.test.js
@@ -1,4 +1,5 @@
 import { compileVariables, excludeDefaultVariableNames, getDefaultVariableNameIndex, getVariables } from "../../../../app/javascript/src/utils/compiler/variables"
+import { completionsMap } from "../../../../app/javascript/src/stores/editor"
 import { disregardWhitespace } from "../../helpers/text"
 
 describe("variables.js", () => {
@@ -12,9 +13,7 @@ describe("variables.js", () => {
         Modify Global Variable(variable5, Test);
         Chase Global Variable At Rate(variable6, Test, ...);
         Chase Global Variable Over Time(variable7, Test, ...);
-        For Global Variable(variable8, 0, 1) {
-          // Do something
-        }
+        For Global Variable(variable8, 0, 1, 1);
       `
       const expectedOutput = {
         globalVariables: ["variable1", "variable2", "variable3", "variable4", "variable5", "variable6", "variable7", "variable8"],
@@ -67,9 +66,7 @@ describe("variables.js", () => {
         Modify Player Variable At Index(Event Player, variable11, Test);
         Chase Player Variable At Rate(Event Player, variable12, Test, ...);
         Chase Player Variable Over Time(Event Player, variable13, Test, ...);
-        For Player Variable(Event Player, variable14, 1) {
-          // Do something
-        }
+        For Player Variable(Event Player, variable14, 0, 1);
       `
       const expectedOutput = {
         globalVariables: [],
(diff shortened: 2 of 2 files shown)


spec/javascript/utils/compiler/variables.test.js
```diff
@@ -397,5 +394,68 @@ describe("variables.js", () => {
           .map((name) => [name, getDefaultVariableNameIndex(name)]))
       ).toStrictEqual(expectedOutput)
     })
+
+    test("Should get Global variables from parameter objects", () => {
+      completionsMap.set(setCompletionsMap("For Global Variable"))
```
netux, line 399: You can do this in a `beforeAll()` block, and it will [set the `completionsMap` before all the `test()`s run](https://jestjs.io/docs/setup-teardown#one-time-setup).

This means you can set all the Workshop Actions you want in the completionsMap at the beginning and not have to worry about making them available in each individual `test()`.

And don't worry, as long as you put the `beforeAll()` block inside the first `describe()` block in the file, your `completionsMap` should persist throughout all the tests (see [Scoping](https://jestjs.io/docs/setup-teardown#scoping) in our test runner docs).

spec/javascript/utils/compiler/variables.test.js
```diff
@@ -397,5 +394,68 @@ describe("variables.js", () => {
...
+    })
   })
 })
+
+function setCompletionsMap(actionLabel) {
+  return [{
+    label: actionLabel,
+    args_length: 4,
+    parameter_keys: ["One", "Two", "Three", "Four"],
+    parameter_defaults: ["A", "B", "C", "D"]
```
netux, line 454: I'd say lets keep these real.

E.g. for `For Global Variable`:
```js
{
  label: "For Global Variable",
  args_length: 4,
  parameter_keys: ["Control Variable", "Range Start", "Range Stop", "Step"]
  parameter_defaults: ["variableName", "0", "Count Of(Array())", "1"]
}
```

E.g. for `Chase Player Variable Over Time`:
```js
{
  label: "Chase Player Variable Over Time",
  args_length: 5,
  parameter_keys: ["Player", "Variable", "Destination", "Rate", "Reevaluation"]
  parameter_defaults: ["Event Player", "variableName", "0", "1", "Destination And Rate"]
}
```

You can get these from autocompletion in the Editor :)

spec/javascript/utils/compiler/variables.test.js
```diff
@@ -397,5 +394,68 @@ describe("variables.js", () => {
...
+        globalVariables: [],
+        playerVariables: ["variable1"]
+      }
+  
+      expect(getVariables(input)).toEqual(expectedOutput)
+    })
   })
 })
+
+function setCompletionsMap(actionLabel) {
```
netux, line 449: This isn't really `set`ting `CompletionsMap`, but rather generating its content. How about a name like "createCompletionMapItems"?

Though I'd guess this function becomes irrelevant if you are moving everything to a `beforeAll()` block.

spec/javascript/utils/compiler/variables.test.js
```diff
@@ -12,9 +13,7 @@ describe("variables.js", () => {
         Modify Global Variable(variable5, Test);
         Chase Global Variable At Rate(variable6, Test, ...);
         Chase Global Variable Over Time(variable7, Test, ...);
-        For Global Variable(variable8, 0, 1) {
-          // Do something
-        }
+        For Global Variable(variable8, 0, 1, 1);
```
netux, line 16: Please don't remove this. I have a feeling it was done intentionally.

spec/javascript/utils/compiler/variables.test.js
```diff
@@ -67,9 +66,7 @@ describe("variables.js", () => {
         Modify Player Variable At Index(Event Player, variable11, Test);
         Chase Player Variable At Rate(Event Player, variable12, Test, ...);
         Chase Player Variable Over Time(Event Player, variable13, Test, ...);
-        For Player Variable(Event Player, variable14, 1) {
-          // Do something
-        }
+        For Player Variable(Event Player, variable14, 0, 1);
```
netux, line 69: Ditto. Please don't edit remove curly braces here. There are in the tests to check that they don't interfere :)

app/javascript/src/utils/compiler/variables.js
```diff
@@ -1,6 +1,7 @@
 import { findRangesOfStrings, getClosingBracket, matchAllOutsideRanges, splitArgumentsString } from "../parse"
+import { evaluateParameterObjects } from "./parameterObjects"
 
-// NOTE: The fact variable names can start with a decimal is intention.
+// NOTE: The fact variable names can start with a decimal is intentional.
```
netux, line 4: Thanks for fixing my typos 😅

netux (changes requested):
Oh, another PR? Sure, then lets close #424.

I think this uses some bad practices for tests/does things in ways that other tests do differently.

spec/javascript/utils/compiler/variables.test.js
Mitcheljager: Better yet, move it to a constant at the start of the file. Writing function in your tests gets "dangerous" as you will need to start writing tests for your test. Write the value to a single constant and when you need to change the label do `constant.label = "new value"`

spec/javascript/utils/compiler/variables.test.js
netux: Except `getVariables()` relies heavily in order of parameters, and also it makes it easier to read the test inputs if they match how they look on real Workshop.

spec/javascript/utils/compiler/variables.test.js
Loottoo: maybe im missing something here but why test that it works when its the wrong syntax? you cant paste it into the game, its not even the right number of parameters

netux: We are not testing that it works in-game, we are testing that it compiles. The compiler is allowed to be dumb sometimes, specially for making tests shorter and more to the point.
Though I guess for this test the curly braces _are_ kind of dumb, especially when there haven't been reports of curly braces after actions breaking `getVariables` before.

I guess my main issue here is that changing this is out of scope for this PR.

spec/javascript/utils/compiler/variables.test.js
Loottoo: isnt that more or less equivalent to what i am already doing, except im setting it in a function? the only thing im changing is the label, the rest is constant.
but yeah the name is not good, i was actually going to change it but i forgor 💀

spec/javascript/utils/compiler/variables.test.js
Loottoo: yeah fair enough, i just assumed it was a mistake but thats probably not something i should just assume 🙃

Mitcheljager: Ah yes, that was definitely a mistake :D

spec/javascript/utils/compiler/variables.test.js
netux: There is `beforeAll()`, which runs before all `test()`s under the same scope, and `beforeEach()`, which runs before each `test()` under the same scope.

I think its better because you can keep things nice and tidy. One part is setup (`beforeAll()`), and the other part is testing and verifying (all the `test()`s).
You can also say "the library provides it and is a common pattern, so it must be good"™️.

spec/javascript/utils/compiler/variables.test.js
```diff
@@ -397,5 +394,86 @@ describe("variables.js", () => {
...
+        args_length: 5,
+        parameter_keys: ["Player", "Variable", "Destination", "Duration", "Reevaluation"],
+        parameter_defaults: ["Event Player", "variableName", "0", "1", "Destination And Duration"]
+      }, {
+        label: "Create Effect",
+        args_length: 5,
+        parameter_keys: ["Visible To", "Type", "Color", "Position", "Radius", "Reevaluation"],
+        parameter_defaults: ["All Players(All Teams)", "Sphere", "Color(white)", "Vector(0, 0, 0)", "1", "Visible To Position And Radius"]
+      }])
+    })
```
netux, line 421: Since this applies to all `test()`s, I would put it at the beginning of the `describe()` block.

netux (approved):
LGTM

spec/javascript/utils/compiler/variables.test.js
Loottoo: Oh i thought id put it where its the most relevant, wasn't aware `beforeAll()` is asynchronous . I agree it makes sense putting it in the beginning then :OllieCrunch: 🦦 🍟

netux: Yeah. If you want to make it run only for certain tests you can wrap it and the `test()`s on another `describe()` block inside the `describe("getVariables")` :D 

But that's not really necessary 🦦🧢🔧

netux:
Oh yeah, good idea on moving the tests to the correct `describe()` 🦦🍝

PR #430, opened 2024-03-05
fix(editor): Fix incorrect compiled values when using Autocomplete using Parameter Objects

Mitcheljager:
When Autocomplete using Parameter Objects was enabled the compiler would include the parameter names in the compiled result, resulting in something like:
```
Create HUD Text(
	 Visible To: All Players(All Teams), Custom String("Test"), 
	 Subheader: Null, 
	 Text: Null, 
	 Location: Left, 
	 Sort Order: 0, 
	 Header Color: Color(white), 
	 Subheader Color: Color(white), 
	 Text Color: Color(white), 
	 Reevaluation: Visible To And String, 
	 Spectators: Default Visibility);
```
This happened because the value set in the completions map for `parameter_default` was set to the same value as `apply`, which is used for autocompletions. As a result it would insert the value as if it was the autocomplete value, including all the keys for the parameter object.

I fixed this by setting the param `parameter_default` _before_ transforming the `apply` value for Parameter Objects.

The code was a bit hard to read before, and this didn't really fix it, but I included some comments and moved a function to hopefully help in the future.

### DIFF
--- a/app/javascript/src/components/editor/Editor.svelte
+++ b/app/javascript/src/components/editor/Editor.svelte
@@ -93,7 +93,7 @@
       const useNewlines = params.args_length >= $settings["autocomplete-min-parameter-newlines"]
 
       // Generate Apply map to be used for autocomplete and other bits
-      let apply = v.args.map(a => {
+      const apply = v.args.map(a => {
         const name = toCapitalize(a.name?.toString().toLowerCase())
         let defaultValue = a.default?.toString().toLowerCase().replaceAll(",", "")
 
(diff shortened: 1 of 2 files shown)


app/javascript/src/utils/text.js
```diff
@@ -0,0 +1,3 @@
+export function toCapitalize(string) {
+  return string.toLowerCase().replace(/(^\w{1})|(\s+\w{1})/g, letter => letter.toUpperCase())
```
netux, line 2: While we are here...

```suggestion
  return string.toLowerCase().replace(/(?<=^|\s)\w/g, letter => letter.toUpperCase())
```

This takes less steps 
previous regex (https://regex101.com/r/Mlbicn/1)
new regex (https://regex101.com/r/Mlbicn/2)

app/javascript/src/components/editor/Editor.svelte
```diff
@@ -91,37 +92,42 @@
...
       params.parameter_keys = detail
-      params.parameter_defaults = apply
+      params.parameter_defaults = apply.map(([_, defaultValue]) => defaultValue)
+
+      // If useParameterObject is enabled transform the apply to include the parameter name.
+      // It's important this happens after setting the parameter_defaults param, as that uses
+      // a different format and we don't want it to use the parameter object format.
+      if (useParameterObject) apply = apply.map(([name, defaultValue]) => ([name, `${ useNewlines ? "\n\t" : "" } ${ name }: ${ defaultValue }`]))
 
+      const applyValues = apply.map(([_, defaultValue]) => defaultValue)
```
netux, line 115: We can avoid modifying `apply`, since it isn't used afterward 😁

```suggestion
      const applyValues = apply.map(([name, defaultValue]) => {
        // If useParameterObject is enabled transform the apply to include the parameter name.
        // It's important this happens after setting the parameter_defaults param, as that uses
        // a different format and we don't want it to use the parameter object format.
        if (useParameterObject) return `${ useNewlines ? "\n\t" : "" } ${ name }: ${ defaultValue }`

        return defaultValue
      })
```

app/javascript/src/utils/text.js
Mitcheljager: Hmm, fair enough, but I will keep it out of this PR

app/javascript/src/components/editor/Editor.svelte
Mitcheljager: Ah sure, that seems like a better solution

app/javascript/src/utils/text.js
netux: I didn't realize I had the flavor set to PHP. Here are some results with JS:

previous regex (https://regex101.com/r/Mlbicn/3)
new regex (https://regex101.com/r/Mlbicn/4)

Doesn't change anything, but I wanted to keep the record straight :P

PR #431, opened 2024-03-09
Mobile editor and DragHandle tweaks

Loottoo:
Added a short delay when touching the EditorList for mobile so that you can scroll and click them easier without accidentally moving the items. (fallbackTolerance should also help)

Moved the settings modal a bit to the right for small screens so that it doesnt get cutoff on the left side.

The requestAnimationFrame in DragHandle.svelte make moving the DragHandles less laggy, should work for both PC and mobile.

### DIFF
--- a/app/assets/stylesheets/elements/_dropdown.scss
+++ b/app/assets/stylesheets/elements/_dropdown.scss
@@ -38,10 +38,6 @@
   overflow-y: auto;
   z-index: 100;
 
-  @media (max-width: 999px) {
-    right: -70px;
-  }
-
   @include styled-scrollbar();
 
   &--left {
--- a/app/assets/stylesheets/elements/_dropdown.scss
+++ b/app/assets/stylesheets/elements/_dropdown.scss
@@ -38,6 +38,10 @@
   overflow-y: auto;
   z-index: 100;
 
+  @media (max-width: 999px) {
+    right: -70px;
+  }
+
   @include styled-scrollbar();
 
   &--left {
--- a/app/javascript/src/components/editor/DragHandle.svelte
+++ b/app/javascript/src/components/editor/DragHandle.svelte
@@ -28,14 +28,19 @@
   function mouseMove(event) {
     if (!isDragging) return
 
-    const difference = (event[direction] || event.targetTouches[0]?.[direction]) - startPosition
-    currentSize = startSize + difference * (align == "right" ? 1 : -1)
+    requestAnimationFrame(() => {
+      const difference = (event[direction] || event.targetTouches[0]?.[direction]) - startPosition
+      currentSize = startSize + difference * (align == "right" ? 1 : -1)
 
-    setCssVariable(key, `${ currentSize }px`)
+      setCssVariable(key, `${ currentSize }px`)
 
-    currentSize = Math.min(Math.max(currentSize, (Math.max(window.innerWidth, window.innerHeight) / 10)), Math.max(window.innerWidth, window.innerHeight) / 2)
+      currentSize = Math.min(
+        Math.max(currentSize, Math.max(window.innerWidth, window.innerHeight) / 10),
+        Math.max(window.innerWidth, window.innerHeight) / 2
+      )
 
-    localStorage.setItem(key, currentSize)
+      localStorage.setItem(key, currentSize)
+    })
   }
 
   function mouseUp() {
--- a/app/javascript/src/components/editor/DragHandle.svelte
+++ b/app/javascript/src/components/editor/DragHandle.svelte
@@ -25,22 +25,21 @@
     startSize = parseInt(currentSize)
   }
 
-  function mouseMove(event) {
+  async function mouseMove(event) {
     if (!isDragging) return
 
-    requestAnimationFrame(() => {
-      const difference = (event[direction] || event.targetTouches[0]?.[direction]) - startPosition
-      currentSize = startSize + difference * (align == "right" ? 1 : -1)
+    await new Promise(res => requestAnimationFrame(res))
+    const difference = (event[direction] || event.targetTouches[0]?.[direction]) - startPosition
+    currentSize = startSize + difference * (align == "right" ? 1 : -1)
 
-      setCssVariable(key, `${ currentSize }px`)
+    setCssVariable(key, `${ currentSize }px`)
 
-      currentSize = Math.min(
-        Math.max(currentSize, Math.max(window.innerWidth, window.innerHeight) / 10),
-        Math.max(window.innerWidth, window.innerHeight) / 2
-      )
+    currentSize = Math.min(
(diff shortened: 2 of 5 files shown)


app/javascript/src/components/editor/DragHandle.svelte
```diff
@@ -28,14 +28,19 @@
   function mouseMove(event) {
     if (!isDragging) return
 
-    const difference = (event[direction] || event.targetTouches[0]?.[direction]) - startPosition
-    currentSize = startSize + difference * (align == "right" ? 1 : -1)
+    requestAnimationFrame(() => {
```
Mitcheljager, line 31: Perhaps we can save an indent with a promise. GitHub won't let me do a full suggestion, but that would be somehting like this.

```suggestion
    await new Promise(res => requestAnimationFrame(res))
```
or
```suggestion
    await new Promise(res => setTimeout(res))
```

Mitcheljager (changes requested):
Good idea! I did have 2 small formatting suggestions

app/assets/stylesheets/elements/_dropdown.scss
```diff
@@ -38,6 +38,10 @@
   overflow-y: auto;
   z-index: 100;
 
+  @media (max-width: 999px) {
+    right: -70px;
+  }
```
Mitcheljager, line 43: Let's keep this mobile first and use the media-min mixin to keep things a little more consistent.

```suggestion
  overflow-y: auto;
  z-index: 100;
  right: -70px;

  @include media-min(mbl) {
    right: 0px;
  }
```

app/assets/stylesheets/elements/_dropdown.scss
```diff
@@ -38,6 +38,10 @@
   overflow-y: auto;
   z-index: 100;
 
+  @media (max-width: 999px) {
+    right: -70px;
```
netux, line 42: This breaks other dropdowns though
image

app/assets/stylesheets/elements/_dropdown.scss
Loottoo: `@media (max-width: 999px)` is already used in another place in this file, adding a `@include media-min(mbl)` seems less consistent to me. i take it the mixin is used in the other .scss files dealing with PC/mobile differences?

app/assets/stylesheets/elements/_dropdown.scss
Loottoo: ah good catch. i thought i checked all the dropdowns but i forgot the save dropdown 😩.
should i just add a new class like this? what would be a good name for it?

```css
.dropdown__adjust-right{
  @media (max-width: 999px) {
    right: -70px;
  }
}
```

app/javascript/src/components/editor/DragHandle.svelte
Loottoo: Sure!

netux: ☝️🤓 Note that, because of the way browsers usually execute the event loop (sync code → async code → request animation frame → timeouts → event handling → repeat [1]), making this a promise may make it run a little bit later than you want.

It probably doesn't affect much here anyways.

[1]: According to MS Copilot, which gave me [this link](https://dev.to/opensign/a-comprehensive-guide-to-the-event-loop-and-asynchronous-programming-in-javascript-46l0) as source. I only remember watching a talk about this from years ago, so I don't know how true it is nowadays.

app/javascript/src/components/editor/EditorList.svelte
```diff
@@ -25,6 +25,9 @@
       swapTreshhold: 0.25,
       multiDrag: true,
       multiDragKey: "ctrl",
+      fallbackTolerance: 3,
+      delayOnTouchOnly: true,
+      delay: 100,
```
netux, line 30: Hmm, for some reason moving the selected item makes it jump straight to the bottom. If you select another item, it fixes itself. Tested with device emulation both on and off:

https://github.com/Mitcheljager/workshop.codes/assets/6181929/a00caa67-3888-4a41-a94a-88e65df70851

I have a feeling it is a bug with sortablejs though 🤔

app/javascript/src/stores/editor.js
```diff
@@ -10,7 +10,7 @@ import { debounced } from "../utils/debounceStore"
 const VARIABLE_EXTRACTION_DEBOUNCE_MS = 500
 
 export const screenWidth = writable(0)
-export const isMobile = derived(screenWidth, $screenWidth => $screenWidth && $screenWidth < 1100)
+export const isMobile = derived(screenWidth, $screenWidth => $screenWidth && $screenWidth < 999)
```
netux: There is this state in-between 1000 and 1100 pixels wide where the Editor sections are sorted horizontally, but you have to drag them vertically. Any clue why that might be?

app/javascript/src/components/editor/EditorList.svelte
Loottoo: Weird 😮 i just tested it on ws.c and its the same thing so it doesnt appear to be caused by this change. 
although im not experiencing any issues without device emulation, only with mobile (same thing after applying these changes, so it seems unrelated).

app/javascript/src/stores/editor.js
Loottoo: i know that the drag-direction is set in DragHandle.svelte with `$: direction = $isMobile ? "pageY" : "pageX"`, and isMobile begins at 1100px. the question is where the editor sections become horizontal at 1000px. They really should match the same width.

app/javascript/src/components/editor/EditorList.svelte
netux: Hmm. I tested it on prod and also in the master branch and it wasn't happening. Then I guess it's a problem for another day.

app/javascript/src/stores/editor.js
netux: > and isMobile begins at 1100px

Doesn't this change make it start at 1000px, not 1100px?

Mitcheljager: We can change $isMobile to start at 1000 to match the styling

Loottoo: yeah thats what i intended, but are you sure it wont change any other interactions expecting 1100? or maybe thats the only 1100

Mitcheljager: This store is only used in the editor. Everything else relies on the media query, which is already 1000. So it was probably an unintentional mismatch. At worst the 1000-1100 range would be less optimal in terms of UX, but we will see.

app/javascript/src/components/editor/EditorList.svelte
netux: I used Brave, which is Chromium-based. Both times on PC.

Regardless, I've made a ticket for this #432, so we can continue the conversation there.

PR #433, opened 2024-03-16
Avoid extracting variables from actions in the Editor if the action doesn't have enough arguments

netux:
A follow up to #424 which should prevent us from adding `undefined` variables to autocomplete.


Mitcheljager (approved):
Makes sense, nice!

PR #438, opened 2024-03-23
feat: Use Redis as cache store

Mitcheljager:
The server is constantly running out of memory. We have a tiny 512MB of memory and we're always just above that, hovering between 550-650mb most often, sometimes even hitting 750MB. One cause of this is cache, which is saved in memory. I've had to upgrade the server couple times, bringing the monthly cost of the dyno from 7 euro to 50 euro, which is quite a big leap.

This PR is a test to see if storing it in a Redis DB will have better results. I'm using Upstash, which costs 20 cents per 100k commands. One commands being one cache read or write. When a user visits the homepage they might request 20 commands. So with that in mind it might not end up being more cost effective, but it's worth a try. I will give it a go for a day and calculate from there how much it will cost. Anything under 10 euro a month I will consider a success. Between 10 and 20 euro would be ok, depending on how much it really reduces the memory usage of the Heroku dyno.


Mitcheljager:
Merging right away as changes are quite minimal, but feel free to comment

Mitcheljager:
Update 1 hour later: Not viable, costs about 15 cents an hour or over 100 euro a month

Loottoo:
what cache is stored on the server? perhaps it could be compressed?

Mitcheljager:
All cache is, I don't think compression is viable. But as it turns out not _that_ much of the memory used is from cache anyway. 50mb at most. Which is enough to try and do something about, but not enough to make the difference.

Issue #436, opened 2024-03-23
[Feature Request] [Editor] Add "Disabled" indicator to Editor when the current file is disabled/hidden


netux:
Just some sort of visual way to let the user know the file they are currently editing is disabled. Could be a little pill on the top right, a dismissable banner at the top, a color-scheme change, or something of the sort

Mitcheljager:
Good idea!

PR #439, opened 2024-03-24
Add a "Hidden" indicator on the editor content area when the current item will be excluded from the output

netux:
demo

on hover

Also:
- Factor out eye icon in <EditorItemHide> into its own component
- Add CSS variable for secondary color (taken from `.button--secondary`)
- Add `_align.scss` with some classes to change `vertical-align`


netux:
Resolved #436

app/javascript/src/components/editor/Editor.svelte
```diff
@@ -28,8 +29,14 @@
...
-  $: if ($currentProject && $sortedItems?.length && $currentItem && !Object.keys($currentItem).length)
-    $currentItem = $sortedItems.filter(i => i.type == "item")?.[0] || {}
+  let isCurrentItemInherentlyHidden = false
+
+  $: if ($currentProject && $sortedItems?.length) {
+    if ($currentItem && !Object.keys($currentItem).length)
+      $currentItem = $sortedItems.filter(i => i.type == "item")?.[0] || {}
+
+    isCurrentItemInherentlyHidden = isInherentlyHidden($currentItem)
+  }
```
Mitcheljager, line 39: I would prefer to see these as separate statements to keep the logic separate, would also prevent you from having to declare the variable separately

```suggestion
  $: if ($currentProject && $sortedItems?.length && $currentItem && !Object.keys($currentItem).length)
  let isCurrentItemInherentlyHidden = false
    $currentItem = $sortedItems.filter(i => i.type == "item")?.[0] || {}

  $: isCurrentItemInherentlyHidden = $currentItem && isInherentlyHidden($currentItem)
```

app/javascript/src/components/editor/Editor.svelte
```diff
@@ -146,6 +153,22 @@
...
+    if (item.parent) {
+      const parent = $sortedItems.find((parentItem) => parentItem.id === item.parent)
+
+      if (parent) {
+        return isInherentlyHidden(parent)
+      }
+    }
+
+    return false
+  }
```
Mitcheljager, line 170: I think we can do with a few fewer indents

```suggestion
  function isInherentlyHidden(item) {
    if (item.hidden) return true
    if (!item.parent) return false
    
    const parent = $sortedItems.find((parentItem) => parentItem.id === item.parent)
    if (!parent) return false
    
    return isInherentlyHidden(parent)
  }
```

app/javascript/src/components/icon/Eye.svelte
```diff
@@ -0,0 +1,13 @@
+<script>
+  export let crossed = false
+</script>
+
+<svg viewBox="0 0 24 24" lns="http://www.w3.org/2000/svg" {... $$restProps}>
```
Mitcheljager, line 5:
```suggestion
<svg viewBox="0 0 24 24" lns="http://www.w3.org/2000/svg" {...$$restProps}>
```

app/assets/stylesheets/editor/_editor.scss
```diff
@@ -123,6 +124,54 @@ $top-height: 4rem;
...
+  }
+
+  &:hover {
+    opacity: 1;
+  }
+
+  &:hover &__tooltip {
+    display: initial;
+  }
+}
```
Mitcheljager, line 173: Using &__ makes it a little hard to search for, I would prefer them separated

```suggestion
.editor-hidden-item-indicator {
  position: absolute;
  top: 1rem;
  right: 1rem;
  padding: 0.125rem 0.5rem;
  font-size: 0.75rem;
  border-radius: 1rem;
  background-color: var(--secondary);
  color: #fff;
  opacity: 0.5;
  transition: opacity 200ms linear;
  user-select: none;
  
  &:hover {
    opacity: 1;
  }
}

.editor-hidden-item-indicator __tooltip {
  $tooltip-bg-color: $bg-darker;
  $tooltip-arrow-size: 0.5rem;

  position: absolute;
  top: calc(100% + #{$tooltip-arrow-size});
  right: 0;
  width: max-content;
  max-width: 200px;
  padding: 0.25em 0.5em;
  background: $tooltip-bg-color;
  pointer-events: none;
  display: none;
  
  .editor-hidden-item-indicator __tooltip:hover & {
    display: initial;
  }

  &::before {
    content: "";
    position: absolute;
    width: 0;
    height: 0;
    top: -$tooltip-arrow-size;
    right: 0.5rem;
    border-inline: $tooltip-arrow-size solid transparent;
    border-bottom: $tooltip-arrow-size solid $tooltip-bg-color;
  }
}
```

Mitcheljager (changes requested):
Looks nice, good feature. I did have a couple comments on some of the structure

app/javascript/src/components/editor/Editor.svelte
netux: That looks a little weird, but I assume you meant this
```suggestion
  $: if ($currentProject && $sortedItems?.length && $currentItem && !Object.keys($currentItem).length)
    $currentItem = $sortedItems.filter(i => i.type == "item")?.[0] || {}

  let isCurrentItemInherentlyHidden = false
  $: isCurrentItemInherentlyHidden = $currentItem && isInherentlyHidden($currentItem)
```

Mitcheljager: No, you don't need to declare the variable separately

app/assets/stylesheets/editor/_editor.scss
netux: Alright. Is `&--` okay though? I see one right below this, in `.drag-handle`

Mitcheljager: Yeah, I do use &--, which is a bit inconsistent, but it's convenience when adding it to stacks like `&:hover, &:active, &--active`

netux: Did it myself instead of applying the suggestion because the suggestion seemed a little off. Lmk if I missed something.

Mitcheljager: That will do!

PR #441, opened 2024-03-25
Add Code Actions, "Transform parameter objects into positional parameters" and vice-versa code action, to the Editor

### LINKED ISSUE #437: [Feature Request] [Editor] Action to transform a workshop action from using positional parameters to parameter objects, and vice versa
`Action(param1, param2, ...)` ↔ `Action({ Parameter 1: param1, Parameter 2: param2, ... })`

We could take a page out of VSCode's book and add a little 💡 icon with a list of actions depending on where the cursor is https://code.visualstudio.com/docs/editor/refactoring

netux:
## Code Actions

Similar to VSCode's code actions, add a lighbulb icon near the user's cursor that can be clicked to select refactor actions[1] in the code

Lightbulb
Lightbulb dropdown open

[1]: In reality, these are not limited to code refactors, but are meant for it anyway

## Transform parameter objects into positional parameters (and vice-versa) code action provider

As the name implies, they convert action arguments to be in one form or the other. E.g. `{ Param 1: value1, Param 2: value2 }` → `(value1, value2)`.

Takes into consideration parameter order and "autocomplete-min-parameter-newlines" for multi-line parameter objects

https://github.com/Mitcheljager/workshop.codes/assets/6181929/932eb9ce-b257-4087-8474-66a30eeb4263

Fixes #437

### DIFF
--- a/app/assets/stylesheets/elements/_codemirror.scss
+++ b/app/assets/stylesheets/elements/_codemirror.scss
@@ -303,4 +303,27 @@
       color: $text-color-light;
     }
   }
+
+  .cm-codeActionsLightbulb {
+    position: absolute;
+    aspect-ratio: 1;
+    line-height: 1;
+    font-size: 0.75rem;
+    background-color: $bg-darker;
+    padding: 0.25rem;
+    border-radius: 2px;
+    cursor: pointer;
+    transform: translate(-100%, -50%);
+    opacity: 0.5;
+    display: inline-block;
+  }
+
+  .cm-codeActionsLightbulb__dropdown {
+    position: absolute;
+    width: 100%;
+    height: 100%;
+    top: 0;
+    left: 0;
+    opacity: 0;
+  }
 }
--- a/app/assets/stylesheets/elements/_codemirror.scss
+++ b/app/assets/stylesheets/elements/_codemirror.scss
@@ -304,7 +304,7 @@
     }
   }
 
-  .cm-codeActionsLightbulb {
+  .code-actions-lightbulb {
     position: absolute;
     aspect-ratio: 1;
     line-height: 1;
@@ -318,7 +318,7 @@
     display: inline-block;
   }
 
-  .cm-codeActionsLightbulb__dropdown {
+  .code-actions-lightbulb__dropdown {
     position: absolute;
     width: 100%;
     height: 100%;
--- a/app/assets/stylesheets/elements/_codemirror.scss
+++ b/app/assets/stylesheets/elements/_codemirror.scss
@@ -313,7 +313,7 @@
     padding: 0.25rem;
     border-radius: 2px;
     cursor: pointer;
-    transform: translate(-100%, -50%);
+    transform: translateX(-100%) translateY(-50%);
     opacity: 0.5;
     display: inline-block;
   }
--- a/app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
+++ b/app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
@@ -36,21 +36,21 @@ export function transformParameterObjectsIntoPositionalParameters({ view, state 
   // empty positional parameters
   if (fileWideParenTo - fileWideParenFrom <= "()".length) return
 
-  const openParameterObjectBracketIndex = actionCallCont.slice(openParenIndex).match(/(?<=^\s*)\{/)?.index ?? -1
-  if (actionCallCont.slice(1).trimStart()[0] === "{") {
+  const openParameterObjectBracketIndex = actionCallCont.slice(openParenIndex + 1).match(/(?<=^\s*)\{/)?.index ?? -1
+  if (openParameterObjectBracketIndex >= 0) {
     // parameter object → positional parameters
 
     const closingParameterObjectBracketIndex = getClosingBracket(actionCallCont, "{", "}", 0)
     if (closingParameterObjectBracketIndex < 0) return
 
-    const fileWideParameterObjectFrom = fileWideParenFrom + openParameterObjectBracketIndex
-    const fileWideParameterObjectTo = fileWideParenFrom + closingParameterObjectBracketIndex
+    const fileWideParameterObjectFrom = (fileWideParenFrom + 1) + openParameterObjectBracketIndex
+    const fileWideParameterObjectTo = (fileWideParenFrom + 1) + closingParameterObjectBracketIndex
 
     actions.push({
       label: "Transform object into positional parameters",
       position: fileWideActionNameStart,
       run() {
-        const parameterObject = state.doc.sliceString(fileWideParameterObjectFrom, fileWideParameterObjectTo)
+        const parameterObject = state.doc.sliceString(fileWideParameterObjectFrom + 1, fileWideParameterObjectTo - 1)
         const parameters = parseParameterObjectContent(parameterObject, 0)
 
         const argumentList = completion.parameter_keys
--- a/app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
+++ b/app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
@@ -47,7 +47,7 @@ export function transformParameterObjectsIntoPositionalParameters({ view, state 
     const fileWideParameterObjectTo = (fileWideParenFrom + 1) + closingParameterObjectBracketIndex
 
     actions.push({
-      label: "Transform object into positional parameters",
+      label: "Transform parameter object into normal parameters",
       position: fileWideActionNameStart,
       run() {
         const parameterObject = state.doc.sliceString(fileWideParameterObjectFrom + 1, fileWideParameterObjectTo - 1)
--- a/app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
+++ b/app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
@@ -46,6 +46,9 @@ export function transformParameterObjectsIntoPositionalParameters({ view, state 
     const fileWideParameterObjectFrom = (fileWideParenFrom + 1) + openParameterObjectBracketIndex
     const fileWideParameterObjectTo = (fileWideParenFrom + 1) + closingParameterObjectBracketIndex
 
+    // there is other stuff after the parameter object (that is not white space)
+    if (!(/^\s*$/.test(state.doc.sliceString(fileWideParameterObjectTo, fileWideParenTo - 1)))) return
+
     actions.push({
       label: "Transform parameter object into normal parameters",
       position: fileWideActionNameStart,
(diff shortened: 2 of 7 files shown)


app/assets/stylesheets/elements/_codemirror.scss
```diff
@@ -303,4 +303,27 @@
...
+  .cm-codeActionsLightbulb {
+    position: absolute;
+    aspect-ratio: 1;
+    line-height: 1;
+    font-size: 0.75rem;
+    background-color: $bg-darker;
+    padding: 0.25rem;
+    border-radius: 2px;
+    cursor: pointer;
+    transform: translate(-100%, -50%);
```
Mitcheljager, line 316: Everywhere else we use the longer value
```suggestion
    transform: translateX(-100%) translateY(-50%);
```

app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
```diff
@@ -0,0 +1,110 @@
...
+
+  const openParameterObjectBracketIndex = actionCallCont.slice(openParenIndex + 1).match(/(?<=^\s*)\{/)?.index ?? -1
+  if (openParameterObjectBracketIndex >= 0) {
+    // parameter object → positional parameters
+
+    const closingParameterObjectBracketIndex = getClosingBracket(actionCallCont, "{", "}", 0)
+    if (closingParameterObjectBracketIndex < 0) return
+
+    const fileWideParameterObjectFrom = (fileWideParenFrom + 1) + openParameterObjectBracketIndex
+    const fileWideParameterObjectTo = (fileWideParenFrom + 1) + closingParameterObjectBracketIndex
```
Mitcheljager, line 47: The brackets seem a little silly, but it's either or
```suggestion
    const fileWideParameterObjectFrom = fileWideParenFrom + openParameterObjectBracketIndex + 1
    const fileWideParameterObjectTo = fileWideParenFrom + closingParameterObjectBracketIndex + 1
```

app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
```diff
@@ -0,0 +1,110 @@
...
+
+    actions.push({
+      label: "Transform object into positional parameters",
```
Mitcheljager, line 50: Perhaps we can change the wording to "regular parameters" or "workshop syntax" as "positional parameters" is a bit technical

app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
```diff
@@ -0,0 +1,110 @@
...
+      position: fileWideActionNameStart,
+      run() {
+        const parameterObject = state.doc.sliceString(fileWideParameterObjectFrom + 1, fileWideParameterObjectTo - 1)
```
Mitcheljager, line 53: It seems a bit odd that this also does + 1, which is also added to the variable a couple lines before.  Is that correct?

app/javascript/src/lib/codeActions.js
```diff
@@ -0,0 +1,96 @@
...
+class CodeActionLightBulbWidget extends WidgetType {
+  constructor(actions) {
+    super()
+
+    this.actions = actions
+  }
+
+  toDOM() {
+    const container = document.createElement("span")
+    container.className = "button cm-codeActionsLightbulb"
```
Mitcheljager, line 23: Since we can pick the classname here we might as well pick something nicer, even if it doesn't quite follow codemirror conventions

app/javascript/src/lib/codeActions.js
```diff
@@ -0,0 +1,96 @@
...
+    container.textContent = "💡"
+
+    const dropdown = document.createElement("select")
+    dropdown.className = "cm-codeActionsLightbulb__dropdown"
```
Mitcheljager, line 27: Same here

app/javascript/src/lib/codeActions.js
```diff
@@ -0,0 +1,96 @@
...
+
+    this.actions = actions
+  }
+
+  toDOM() {
+    const container = document.createElement("span")
+    container.className = "button cm-codeActionsLightbulb"
+    container.textContent = "💡"
+
+    const dropdown = document.createElement("select")
```
Mitcheljager, line 26: Why is this a select with options rather than a more regular div with buttons inside?

app/javascript/src/lib/codeActions.js
```diff
@@ -0,0 +1,96 @@
...
+
+        if (actions.length > 0) {
+          const { from } = update.state.selection.main
+          const decorationPosition = actions.reduce((position, action) => Math.min(action.position ?? Infinity, position), from)
+
+          this.decorations = Decoration.set([
+            createLightBulbDecoration(decorationPosition, actions)
+          ])
+        }
+      }
```
Mitcheljager, line 91: We could do with some fewer indents here

```suggestion
      if (!(update.docChanged || update.selectionSet || update.viewportChanged))
      
      this.decorations = Decoration.none

      const actions = providers
        .map((provider) => provider(update) ?? [])
        .flat()

      if (!actions.length) return
      
      const { from } = update.state.selection.main
      const decorationPosition = actions.reduce((position, action) => Math.min(action.position ?? Infinity, position), from)

      this.decorations = Decoration.set([
        createLightBulbDecoration(decorationPosition, actions)
      ])
```

app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
```diff
@@ -0,0 +1,110 @@
...
+      run() {
+        const args = splitArgumentsString(state.doc.sliceString(fileWideParenFrom + 1, fileWideParenTo - 1))
+
+        const isMultiLine = args.length > get(settings)["autocomplete-min-parameter-newlines"]
+        const indent = "\t".repeat(getIndentForLine(state, cursorFrom, cursorFrom))
+
+        let parameterObject = "{"
+        if (isMultiLine) parameterObject += "\n"
+        else parameterObject += " "
+        parameterObject += args
```
Mitcheljager, line 86: Let;s add some new lines, it's a little hard to follow

```suggestion
        let parameterObject = "{"
        
        if (isMultiLine) parameterObject += "\n"
        else parameterObject += " "
        
        parameterObject += args
```

app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
```diff
@@ -0,0 +1,110 @@
...
+          .map((value, index) => {
+            const keyValue = `${ completion.parameter_keys[index] }: ${ value.trim() }`
+            return (isMultiLine ? (indent + "\t") : "") + keyValue
+          })
+          .join(",\n")
+        if (isMultiLine) parameterObject += "\n" + indent
+        else parameterObject += " "
+        parameterObject += "}"
```
Mitcheljager, line 94:
```suggestion
        if (isMultiLine) parameterObject += "\n" + indent
        else parameterObject += " "
        
        parameterObject += "}"
```

Mitcheljager (changes requested):
Very cool feature, nice use of widgets. I left quite a few comments, mostly concerning of formatting.

app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
netux: Thee brackets are there for readability. I wanted these calculations to read like "starting from the opening parenthesis, plus the index of the parameter object opening/closing bracket (which is already relative to that opening parenthesis)"

In my head it makes sense that we have to +1 the `fileWideParenFrom`, since we want to get _past_ that parenthesis.

app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
netux: I was hoping people would learn what positional parameters are along the way, but sure.

"Positional parameters" sounds weird anyway, since normally these would be called "positional arguments"[[1]]. But we already called the object notation "parameter objects" instead of "argument objects", so eh.

[1]: https://stackoverflow.com/a/156787

app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
netux: Yes, the +1 and -1 here are to get the stuff inside the curly braces (the "inner content" of the parameter objects).

app/javascript/src/lib/codeActions.js
netux: I saw we already had some custom stuff using the `cm-` notation and just went with it as well.
But what would you suggest?

app/javascript/src/lib/codeActions.js
netux: Easier to implement 😛. Also better for mobile, I guess.

If it looks too ugly I guess I can try and change it.

app/javascript/src/lib/codeActions.js
netux: 👍. I left the stuff from the example here.

app/javascript/src/lib/codeActionProviders/transformParameterObjectsIntoPositionalParameters.js
netux: Adding an extra newline for consistency

```suggestion

        if (isMultiLine) parameterObject += "\n" + indent
        else parameterObject += " "
        
        parameterObject += "}"
```

app/javascript/src/lib/codeActions.js
Mitcheljager: I think it makes sense to use classnames of our own here and keep `cm-` to codemirror specific things so we can always differentiate between the two

app/javascript/src/lib/codeActions.js
Mitcheljager: Fair enough!

app/javascript/src/lib/codeActions.js
netux: Would it make sense to keep that in codeMirror.scss, or should I create a separate file for the extension?

Mitcheljager: Hmm good question, do what you think is best. That file is a bit messy unfortunately

netux: For now I'll just leave it in `_codemirror.scss`. Splitting that file up can be done in another PR :)

I also went with `.code-action-lightbulb` for the class name.

netux:
Added a quick check to avoid providing the parameter object → positional parameters action if there is some more arguments after the parameter object

example

Mitcheljager (approved):
Nice!

PR #443, opened 2024-03-26
refactor: Unify css and scss variables

Mitcheljager:
This is another next to impossible to review PR because of the sweeping changes.

We were using a mix of scss and css variables. This was initially done to make adding accessibility settings and variables colors easier. But over time it has proven to be an annoyance. You don't know when to use a css var and when to use a scss var. This PR changes that by using scss variables for _everything_, css vars are only still included for people to use in custom css or for SVGs. There is one half-way exception for $primary, which uses var(--primary, $primary) so it can be easily overwritten for the wiki, but in $primary is still the source of truth. 

With this you'll notice that accessibility settings are gone. Over time I've come to realize that separate accessibility settings are not the way to go. Users have to log in for it to be accessible, which is bad. The website should be accessible by default, not by exception.

Everything should look identical to how it did before.


netux (approved):
Yeah... I won't even attempt to parse the changes here.

As long you kept all CSS variables, I'm fine with this :)
Sad to see accessibility features gone. Though browsers should already provide those settings, and we could definitely support them passively through CSS via [`prefers-contrast`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast), [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion), and [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).

Mitcheljager:
@netux Exactly, we should make it accessible through tools that already exist (we're doing pretty good already to be fair)

netux:
Is it possible for `.high-contrast` to come back in the form of `@media (prefers-contrast: more)`? Or are you fully against going back to CSS variables again?

Mitcheljager:
@netux It certainly can! But likely through other means than css variables, or a route similar to what I did for $primary

netux:
Is this too bad?

```scss
$text-color: var(--high-contrast-text-color, $gray);
$text-color-light: var(--high-contrast-text-color-light, lighten($gray, 10%));
$text-color-lightest: var(--high-contrast-text-color-lightest, #dfdfdf);
$text-color-dark: var(--high-contrast-text-color-dark, $gray-dark);
$bg-dark: var(--high-contrast-bg-dark, #2e3337);

:root {
  @media (prefers-contrast: more) {
    --high-contrast-text-color-lightest: #fff;
    --high-contrast-text-color: #{ lighten($gray, 15%) };
    --high-contrast-text-color-light: #{ lighten($gray, 10%) };
    --high-contrast-text-color-dark: #{ $gray };
    --bg-darker: #{ darken($body-bg, 10%) };
  }
}
```

Mitcheljager:
Yes 😅. The main point of going back to scss variables was so we could use `darken`, `lighten`, and such, which isn't possible with css variables (yet!).

Mitcheljager:
@netux I realise I didn't really eloborate there. I think in order to get more contrast where it's needed we can locally add the selectors. Either by doing something like `var(--text-color, $text-color)` as the color, which can be overwritten on a global level. Or, I think preferably, it would be simply with a media query but for each selector. Which is a little more work, but in reality there's only a handful of places where that will be needed. Most of the values we use are well within high contrast ratios and most that aren't can likely be solved fairly easily my making some of the backgrounds a little darker.

PR #445, opened 2024-03-27
Fix Battle.net/Discord Icons & Button/Icon Colors

Js41637:
## Changes
- Removes duplicate Battle.net & Discord icons, there was 2 of each even though they are basically identical?
- Update the Battle.net & Discord icons to match the real logos
- Updates the Battle.net & Discord colors to match the proper brand colors
- The new Battle.net & Discord SVGs allow changing the color to whatever you want with CSS instead of hardcoding it in the SVG
  - This means by default the icons have no color and will appear black unless you manually specify a color
  - Buttons set the text color to white resulting in a white icon
  - CSS classes have been added for setting the icon to either Battle.net's or Discord's color
- Fixes bug on Discord buttons where the Discord icon is the same color as the button background making the icon invisible, now that the icon inherits the text color, the icon is now white

I realise the Discord icon issues were caused by a recent commit: https://github.com/Mitcheljager/workshop.codes/commit/037317e9c6debb95657913759cd20964ff69c2cb, i actually had no idea you only just changed this like an hour ago. I just happened to notice it on the site so if you prefer the older style Discord icon I could update it to continue to use that one if you want.

I think the new discord icon is better for most cases but in the header I think having the background behind it is nice as well, there's also a simple CSS solution for that as well which is just making the icon white and making the background blue with a border radius, basically replicates the original icon.

## Images
| Before  | After  |
|---|---|
| image  | image  |
| image  | image  |
| image  |  image |
| image  |  image |

## Important
I haven't actually tested the changes myself as I did not want to go through all the effort of getting the site running locally. Also because it dealt with login I figured it might be even more of a hassle. 

Images above were made by manually editing the CSS and SVGs on the website and should reflect what the changes do but I have no idea if they technically work. It seems like they should but you will need to vertify to make sure😺


app/assets/stylesheets/utilities/_colors.scss
```diff
@@ -6,6 +6,18 @@
...
+  color: $white;
+}
+
+.icon-discord {
+  color: $discord;
+}
+
+.icon-battlenet {
+  color: $bnet;
+}
```
Mitcheljager, line 19: Let's use text-* instead to be more consistent with the others, I think `text-discord` and `text-white` already exists even

```suggestion
.text-battlenet {
  color: $bnet;
}
```

app/views/application/_header.html.erb
```diff
@@ -10,7 +10,7 @@
       <%= link_to "On Fire", on_fire_path, class: "navigation__item #{ "navigation__item--active" if controller_name == "on_fire" }" %>
       <%= link_to "Wiki", wiki_root_path, class: "navigation__item #{ "navigation__item--active" if is_wiki? }" %>
       <%= link_to "Editor", editor_path, class: "navigation__item hidden visible-lg", target: "_blank" %>
-      <%= link_to inline_svg_tag("discord.svg"), "https://discord.gg/DdxsQRD", target: "_blank", rel: "noreferrer noopener", class: "navigation__item mt-1/4 mbl:mt-1/8", aria: { label: "Join the Workhsop.codes Discord" } %>
+      <%= link_to inline_svg_tag("icons/icon-discord.svg", width: 25, height: 25), "https://discord.gg/DdxsQRD", target: "_blank", rel: "noreferrer noopener", style: "display: flex", class: "navigation__item", aria: { label: "Join the Workhsop.codes Discord" } %>
```
Mitcheljager, line 13: Instead of the inline style, there is a `flex` utility class (why is this class necessary anyway?)

```suggestion
      <%= link_to inline_svg_tag("icons/icon-discord.svg", width: 25, height: 25), "https://discord.gg/DdxsQRD", target: "_blank", rel: "noreferrer noopener", class: "navigation__item flex", aria: { label: "Join the Workhsop.codes Discord" } %>
```

app/views/blocks/profile/_socials.html.erb
```diff
@@ -1,14 +1,14 @@
 <div class="content-block">
   <% if block_property(block, "battlenet").present? %>
     <div class="content-block__social">
-      <%= inline_svg_tag "icons/icon-battlenet.svg", height: 21, width: 21 %>
+      <%= inline_svg_tag "icons/icon-battlenet.svg", height: 21, width: 21, class: "icon-battlenet" %>
```
Mitcheljager, line 4: This would go with the review above
```suggestion
      <%= inline_svg_tag "icons/icon-battlenet.svg", height: 21, width: 21, class: "text-battlenet" %>
```

app/views/blocks/profile/_socials.html.erb
```diff
@@ -1,14 +1,14 @@
...
       <%= block_property(block, "battlenet") %>
     </div>
   <% end %>
 
   <% if block_property(block, "discord").present? %>
     <div class="content-block__social">
-      <%= inline_svg_tag "icons/icon-discord.svg", height: 21, width: 21 %>
+      <%= inline_svg_tag "icons/icon-discord.svg", height: 21, width: 21, class: "icon-discord" %>
```
Mitcheljager, line 11:
```suggestion
      <%= inline_svg_tag "icons/icon-discord.svg", height: 21, width: 21, class: "text-discord" %>
```

Mitcheljager (changes requested):
Nice, good change! I did have a few comments on formatting

app/views/application/_header.html.erb
Js41637: The flexx CSS class also has 100% width for some reason so i couldn't use it. It just centers the icon. The one before used some margins to try and center it but just setting it to flex does it automatically.

Mitcheljager: Ah ok, that's annoying

app/assets/stylesheets/utilities/_colors.scss
Js41637: I was going to but I figured maybe if there was a case where you wanted button specific CSS but it is pretty unlikely so I'll change them to use the text ones.

Mitcheljager (approved):
Nice!

app/views/application/_header.html.erb
Js41637: I updated it to use a class, I forgot i also needed to add the `text-discord` on the icon so I figured it might as well just use its own navigation item class. I'm not 100% sure how it works in practice as the navigation items color can change depending on hover or the page your on and i'm unsure if that would override the color of the icon.

Mitcheljager: I think it would, but that's ok, it needs a hover state after all

Js41637: It already has the underline from the nav item. I think maybe i'll just use `!important` to keep it blue, hoving turns it white at the moment which I think is a bit weird. Though the wiki page shows all the text as white so you could also make the icon white on the wiki page if you wanted.

Mitcheljager: Let's not use !important, that will be annoying in the future. If you want it to stay purple you can overwrite the hover state

```
&:hover {
  color: $discord;
}
```

Js41637:
Ok I think it's done. It does need to be tested quickly though to make sure it actually works as I haven't actually ran the changes locally. 😔

Mitcheljager:
I will merge it in, test it, and fix it in case anything needs fixing

PR #446, opened 2024-04-01
refactor: Replace Webpack with Vite

Mitcheljager:
This PR replaces Webpack with Vite. The PR seems fairly massive, but that's largely because of moved files. Let's go over what this means:

- **All SCSS is now handled through Vite rather than Sprockets**. All SCSS files are moved to `/app/javascript`, which is a fairly confusing folder to find SCSS in and I might end up renaming it (`frontend` is the convention).
  - SCSS now uses HMR.
  - Images used in SCSS file are copy pasted over to the same folder structure, meaning we have duplicate files. This sucks for now, but I plan on moving all images files to this same location so that all images are handled through vite, but that's for another PR.
  - Dividing in SCSS is deprecated, this is because of aspect rations. SCSS would have no way to differentiate between `16 / 9` and `16 / 9`. I used an automatic converter to replace divisions with multiplication where possible or `math.div(1, 2)` where not. I don't love the syntax, but that's ok.
- **JS files are handled through Vite**. This is no different than how it was with Webpacker in where to find the files.
  - The Zez-ui didn't work with it's odd imports. I've replaced the entire source code with a minified version. I might revisit this later, but I didn't want to debug longer, it's deprecated anyway.
  - We now have HMR! This works in all Svelte and JS files. For files of the website itself this seems a little flacky, perhaps to do with how things are imported, with each file being called in the main *.js files. But that's a problem for another day. HMR for the Editor is a blessing. 
- **Jest is replaced with Vitest**. All tests for the editor now use Vitest. The syntax is just about the same, with the exception of having to import vitest functions, which is fine.
- **Javascript in rspec and cucumber wouldn't work**. I couldn't get Javascript in these feature tests to work at all. I spent several hours on this and I've given up for now. I've disabled the few tests that use javascript and I want to revisist them later (perhaps do it through vitest instead?).
- **Node version is now a minimum of 18.12.0**. This is kind of an arbitrary version, it just happened to be what I had installed. You should be good with 20+ too, but I'm not sure.

We can now also use path aliases in javascript imports (No more `../../../../`), but I haven't actually set up any aliases yet, that's for another PR.

To run the vite server during development all you have to do is run `ruby bin/vite dev` just like you'd do `ruby bin/webpack-dev-server` before. I've also updated the `launch.json` file, but since I don't use that myself I'm not certain it works.

I wouldn't quite recommend going through all the changes. If you do want to, go commit by commit. A lot of it was auto generated by the lovely Vite.


app/javascript/src/utils/compiler/variables.js
```diff
@@ -164,8 +164,8 @@ export function getVariables(joinedItems) {
   const literalGlobalVariables = matchAllOutsideRanges(stringRanges, joinedItems, globalVariablesRegex).map((match) => match[0])
   const literalPlayerVariables = getLiteralPlayerVariables(joinedItems, stringRanges)
 
-  const playerVariables = [...new Set([...literalPlayerVariables, ...playerVariablesFromActions])]
-  const globalVariables = [...new Set([...literalGlobalVariables, ...globalVariablesFromActions])]
+  const playerVariables = [...new Set([...literalPlayerVariables, ...playerVariablesFromActions])].filter(v => v)
+  const globalVariables = [...new Set([...literalGlobalVariables, ...globalVariablesFromActions])].filter(v => v)
```
netux, line 168: Typecheck issue?

app/javascript/fonts/zez_ui/dump_chars.py
netux, line 1: Is this script really needed?

.vscode/launch.json
netux, line 1: Changes look good 👍
Though GitHub makes it very apparent that the indentation is all over the place. Could I ask you run a little reindent all lines here while you are at it? 🥺

netux (commented):
Ameizing!
I'll test tomorrow. Here are some preliminary comments.

app/javascript/src/utils/compiler/variables.js
Mitcheljager: No, it actually returned undefined. Not sure why it didn't before. I didn't debug very much

app/javascript/fonts/zez_ui/dump_chars.py
Mitcheljager: No

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "overwatch-workshop",
   "private": true,
   "engines": {
-    "node": "^14.0.0"
+    "node": "^18.12.0"
```
netux, line 5: I think we could be more lenient here

```suggestion
    "node": ">18"
```

app/javascript/src/utils/compiler/variables.js
Mitcheljager: It wasn't a project, the tests were failing. This one specifically: https://github.com/Mitcheljager/workshop.codes/blob/4e67b79c04b8a1496bfaf7a0f15973bc9534155d/app/javascript/test/utils/compiler/variables.test.js#L316-L328

.vscode/launch.json
netux: Ehh, I just did it myself: https://github.com/Mitcheljager/workshop.codes/pull/447

netux (approved):
After upgrading Node to _exactly, [und exactly](https://www.youtube.com/clip/UgkxKytsWZjLqKgPyJ6dwqoLwGWqncNyW-1x)_ 18.12.0, reinstalling yarn (why do we still have yarn v1?), reinstalling all node dependencies and ruby gems, everything seems to work as expected with my VSCode setup. Pretty nice :)

app/javascript/src/utils/compiler/variables.js
netux: That's interesting. I think it may have been a bug with Jest's compare logic (or maybe we were using the wrong function?).

clearly not the same output
(Using Jest here - these objects are _not_ the same)

Vite correctly compares the objects how we want it to, so that's why this issue appeared.
I will fix the underlying issue on another PR.

PR #450, opened 2024-04-02
chore: ESLint improvements

Mitcheljager:
This PR adds more ESLint configurations to make our code more consistent. I'm not going to go through each change. The vast majority of changes you will see are for spaces in interpolated values `${ thing } -> ${thing}`. I wanted to do the same for svelte blocks `{ #each } -> {#each}` but I couldn't find the configuration. Many other rules are rules we already kind of had in place, but now they will be enforced.


netux:
> I wanted to do the same for svelte blocks `{ #each } -> {#each}` but I couldn't find the configuration.

[mustache-spacing](https://sveltejs.github.io/eslint-plugin-svelte/rules/mustache-spacing/)?

Mitcheljager:
@netux That's it!

app/javascript/src/get-snippet.js
```diff
@@ -35,6 +35,6 @@ function loadSnippet(event, element) {
     })
     .catch(error => {
       console.error(error)
-      ideElement.innerHTML = `Failed to load Snippet <br>${error}`
+      ideElement.innerText = `Failed to load Snippet <br />${error}`
```
netux, line 38: Wouldn't this cause the `<br />` to be set as part of the text?

Mitcheljager: Ah yup, of course

PR #453, opened 2024-04-15
feat: Add inline filter to search/filter pages

Mitcheljager:
This PR adds a filter to the search pages. This filter is the same as is available in the search bar filter, but more readily available. Heroes are displayed in a grid and the remaining filters are shown just like they are in the filter dropdown. It uses the exact same partial as the filter dropdown. The partial has to be adjusted to allow for this, as not every option is wanted here. Unlike the dropdown items here navigate to the page directly, rather than when applying the filter.

On mobile the grid of hero icons is shorted by default and it can be expanded by click.

### Filter by default
Desktop | Mobile
--- | ---
image | image

### Filter in action
Desktop | Mobile
--- | ---
filter-desktop | filter-mobile


netux:
I don't know if I'm a big fan of that hero selection. Seems way too prominent and big, and I doubt that many people would use it. 

While cool, it'd probably make more sense to convert the selector back to a multi-select, and maybe add the hero icons to the dropdown items themselves?
image

netux:
I also don't like how the view jumps if there one or more creators matching the search, and the search disappears because a hero was selected from the filter:

https://github.com/Mitcheljager/workshop.codes/assets/6181929/b6475a25-8745-4cad-a072-7f57f75d1fef

app/views/filter/_filter_content.html.erb
```diff
@@ -52,7 +60,7 @@
   <div class="filter__item <%= "filter__item--active" if params[:author] %> dropdown" data-dropdown data-filter data-default="Author...">
     <%= link_to params[:author] ? params[:author].split("#")[0] : "Author...", "#",
                 data: { action: "toggle-dropdown get-verified-users", filter_type: "author", value: (CGI.escape(params[:author]) if params[:author].present?) } %>
-    <%= link_to "X", "#", class: "filter__remove", data: { action: "add-filter", value: "" } %>
+    <%= link_to "X", "#", class: "filter__remove", data: { action: direct ? nil : "add-filter", value: "" } %>
```
netux, line 63:
```suggestion
    <%= link_to "X", build_filter_path(:author, nil), class: "filter__remove", data: { action: direct ? nil : "add-filter", value: "" } %>
```

netux: With this, I noticed selecting an author removes the search term. Is that intentional?

Mitcheljager: Nope!

Mitcheljager:
@netux That's fair, the intent was to fill up this view a bit more. It looks a little empty, and yet it's one of the most used pages. I will use the regular dropdown for now and come up with something different later... maybe

Mitcheljager:
@netux Here's what it looks like now

Default | Hero dropdown
--- | ---
image | image

You'll notice the author filter isn't there at all. Solving the issue you mentioned was going to take a good bit of work and with how little that filter is used, I don't want to bother :D

PR #455, opened 2024-04-16
refactor: Import metric graphics charts async

Mitcheljager:
A large part of the logged in user specific JS is made up of these charts. Most users will not see this as it only shows when looking at analytics. So to reduce the file size of that entry this commit changes the imports to be fetched async when creating the charts.

This reduces logged-in-user.js from 826kb (270kb gzipped) to 369kb (123kb gzipped)


netux (approved):
Good idea!

PR #456, opened 2024-04-29
fix(editor): Prevent splitting each on commas in strings

Mitcheljager:
`@each` would incorrect split on commas inside of quotes. So `["Thing", "Some, thing"]` would be output as `["\"Thing\"", "\"Some,", "thing\""]`

This PR fixes that by checking all the valid commas before parsing the array, and if the given comma is not in the valid commas, skip that iteration.

### DIFF
--- a/app/javascript/src/utils/parse.js
+++ b/app/javascript/src/utils/parse.js
@@ -131,3 +131,19 @@ export function matchAllOutsideRanges(ranges, content, regex) {
   }
   return matches
 }
+
+export function getCommasIndexesOutsideQuotes(string) {
+  const commaIndexes = []
+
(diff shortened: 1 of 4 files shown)


app/javascript/src/utils/parse.js
```diff
@@ -131,3 +131,19 @@ export function matchAllOutsideRanges(ranges, content, regex) {
   }
   return matches
 }
+
+export function getCommasIndexesOutsideQuotes(string) {
```
netux, line 135: This could probably use `findRangesOfStrings()`

Mitcheljager: Ah, sure, forgot that existed

PR #459, opened 2024-05-14
Increase Discord icon visibility in nav bar

netux:
- On the wiki and mobile nav menu, use a white color scheme
    | On the wiki | On the mobile nav menu |
    | --- | --- |
    | on the wiki | on the mobile nav menu |
- Outside of that, simply add a slight white glow that makes the icon pop
    switching glow on and off


Mitcheljager (approved):
I can't say I like the box shadow, but I understand why you did that, so let's keep it in. I might end up removing it later if I really dislike it though :D

PR #461, opened 2024-05-18
Increase Discord icon visibility in nav bar, part 2

netux:
Follow-up to #459 which apparently was broken.

I simplified the stylesheet by using more CSS variables, removed the hover color overwrite in the Wiki, and reduced the drop shadow effect in non-Wiki pages.

### DIFF
--- a/app/javascript/scss/elements/_navigation.scss
+++ b/app/javascript/scss/elements/_navigation.scss
@@ -91,10 +91,6 @@
     &:hover {
       color: $discord;
     }
-
-    .wiki & {
-      color: $white;
-    }
   }
 }
 
--- a/app/javascript/scss/elements/_navigation.scss
+++ b/app/javascript/scss/elements/_navigation.scss
@@ -18,6 +18,8 @@
 
 .navigation__item {
   --color: #{$text-color-lightest};
+  --discord-color: #{$discord};
+  --discord-drop-shadow: drop-shadow(0 0 4px #{$white});
   display: block;
   margin-bottom: $margin * 0.25;
   border-radius: $border-radius * 0.5;
@@ -67,6 +69,8 @@
 
   .wiki & {
     --color: #{$white};
+    --discord-color: #{$white};
+    --discord-drop-shadow: none;
   }
 
   svg {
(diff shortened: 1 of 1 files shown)


app/javascript/scss/elements/_navigation.scss
```diff
@@ -18,6 +20,8 @@
 
 .navigation__item {
   --color: #{$text-color-lightest};
+  --discord-color: #{$discord};
+  --discord-drop-shadow: drop-shadow(0 0 4px #{color.change($white, $alpha: 0.2)});
```
Mitcheljager, line 24: You can just use rgba instead of the whole `sass:color` thing. 
```suggestion
  --discord-drop-shadow: drop-shadow(0 0 4px rgba(#{$white}, 0.2));
```

Mitcheljager (changes requested):

app/javascript/scss/elements/_navigation.scss
netux: This doesn't work because `$white` is in hex format
demo in MDN throws an error

netux: This does seem to work, but is not supported in Firefox[1]: `rgba(from #{$white} r g b / 0.2)`

[1]: [`rgb()` Browser Compatibility](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb#browser_compatibility), Firefox doesn't support Relative RGB colors

Mitcheljager: I know, but SCSS should transform it just fine. It's how we use it everywhere. This line for example:
https://github.com/Mitcheljager/workshop.codes/blob/6cd52cc0303345e3070ed231ed232f9eef198fd5/app/javascript/scss/elements/_dropdown.scss#L38

Perhaps it doesn't work inside of a CSS variable since it's an SCSS function. Did you try `#{rgba($white, 0.2)}`?

netux: That works!

Mitcheljager: Nice!

Mitcheljager (approved):

PR #462, opened 2024-05-18
fix(editor): increase text-indent multiplier for word-wrapping indent in Chromium browsers

netux:
Turns out Chrome and Firefox handle negative indents differently. This fixes the duplicate indents in Chrome, while keeping the indents normal for Firefox.

~~This code assumes there are only two browser engines. Sorry Ladybird :(~~


Mitcheljager (approved):
What, that's awfully weird

PR #484, opened 2024-10-01
fix: pr comments on #480

CactusPuppy:
This PR addresses the review comments on #480

- **style: make class name more BEM-compliant**
- **fix(editor): prevent folder arrow indicator from appearing above Add Item button**


Mitcheljager (approved):
Neat!

PR #489, opened 2024-11-09
refactor: Use typescript for (almost) all main site javascript

Mitcheljager:
I was going to implement Typescript specifically for the editor, but I decided to get it started for the main site javascript first. I did everything in one big go, which may have been a mistake. There were quite a few files and I didn't spend super long on any of them, as a result I probably took some shortcuts I shouldn't have, but at least we now have type safety.

I skipped the markdown editor scripts as they are pretty complex and I want to take care of them separately.

In the end I'm not entirely sure it's really worth it, these utils aren't very complex and a lot of the type safety here doesn't do much. I've had to put in a lot of early returns in case of missing elements, but in reality these elements will never be missing.

I don't expect anyone to review this, as it's a rather nasty PR. I will early merge it soon-ish or use it as a base PR when tackling the editor compiler. Svelte components will have to wait until the Svelte 5 upgrade.


netux:
> I've had to put in a lot of early returns in case of missing elements, but in reality these elements will never be missing.

You can use non-null assertion operation (`!`), i.e.
```ts
something // type: whatever | null | undefined
something! // type: whatever
```

PR #493, opened 2024-11-19
swapped ability 1 and 2 icon file names

tweci:
switched ability icon names for console platforms as they were reversed


Mitcheljager (approved):
Whoops, thanks!

PR #497, opened 2024-11-26
feat: Upgrade value/action wiki articles with rake task

Mitcheljager:
This PR adds a rake task to upgrade wiki articles. This rake task is as ugly as it gets, but as it's a one time thing, I don't care. Verboseness over cleverness, and temporary cleverness over maintainability here.

Articles are changed to a new format;
- Description
- Snippet
- Properties
  - Return value
  - Parameters
    - Parameter details
    
Each article will now have a tiny snippet, this is super basic and only shows the value/action with it's default values. In some cases it doesn't make sense to use these in their isolated context, but it helps with copy pasting.

Previously the articles had a separate "return" section, but this is now under the "properties" section. This section lists both the return value and the parameters. The parameters are shown with a large details block, similar to before, with all their details given. In these blocks types are automatically given a link to their corresponding article if it exists. This is usually (always?) to constants like "HUD Reevaluation", "Color", "Button", etc.

The description is a little unique. Some articles have had their descriptions edited over time and I don't want to override those. If the description of the article is not the same as that in the yml file, we use the already existing description. If they are the same a new description is inserted with some new formatting.

Before | After
--- | ---
image |image
image | image
image | image
image | image
image | image
image | image
image | image

### DIFF
--- a/lib/tasks/upgrade_wiki_articles.rake
+++ b/lib/tasks/upgrade_wiki_articles.rake
@@ -95,7 +95,3 @@ task :upgrade_wiki_articles => :environment do
     )
   end
 end
-
-def to_formatted_slug(string)
-  string.gsub(/[^0-9a-z ]/i, "").parameterize.gsub("+", "-").downcase
-end
--- a/lib/tasks/upgrade_wiki_articles.rake
+++ b/lib/tasks/upgrade_wiki_articles.rake
@@ -1,0 +1,101 @@
+desc "Upgrade wiki articles from old formatting to somewhat fancier formatting"
+task :upgrade_wiki_articles => :environment do
+  actions = YAML.safe_load(File.read(Rails.root.join("config/arrays/wiki", "actions.yml")))
+  values = YAML.safe_load(File.read(Rails.root.join("config/arrays/wiki", "values.yml")))
+  defaults = YAML.safe_load(File.read(Rails.root.join("config/arrays/wiki", "defaults.yml")))
+
+  user = User.find_by_username("mitsiee")
+  user = User.find_by_username("admin") unless @user.present?
+
+  (actions + values).each do |current|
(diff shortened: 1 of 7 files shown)


netux:
> If the description of the article is not the same as that in the yml file, we use the already existing description

Did you also check for arguments getting changes?

lib/tasks/upgrade_wiki_articles.rake
```diff
@@ -0,0 +1,101 @@
...
+
+#{ description_only == current["description"] ? "<big>#{ description_only }</big>" : description_only }
+
+## Snippet
+
+```
+#{ current["en-US"] }#{ current["args"].present? ? parameters_snippet : "" };
+```
+
+## Properties
```
netux, line 67:
```suggestion
## Arguments
```

netux: or

```suggestion
## Parameters
```

(matching what workshop.codes/editor calls them)

Mitcheljager:
@netux Nope, those are intentionally overwritten.

lib/tasks/upgrade_wiki_articles.rake
Mitcheljager: This is the general header for the properties of the action, which includes both the return value and parameters

netux: Ah I see. Didn't scroll further down 👍

Mitcheljager: I went with parameters over arguments. They are used interchangeably, but they don't actually mean the same thing. To put it in Javascript:

```
function thing(one, two) { ... } <- Parameters

thing(one, two) <- Arguments
```

The available options in the function are parameters, but you give arguments as values for those parameters.

PR #518, opened 2025-03-10
fix(editor compiler): correctly gen erate the arguments for the Custom String() replacing a @translate directive when compiling a singleLanguageOverride

netux:
Very wordy title just to say `@translate("key", "a", "b", "c")` was being compiled to `Custom String("value of key")` without `"a"`, `"b"`, nor `"c"`.

### DIFF
--- a/app/javascript/src/utils/compiler/translations.ts
+++ b/app/javascript/src/utils/compiler/translations.ts
@@ -24,16 +24,15 @@ export function convertTranslations(joinedItems: string, singleLanguageOverride:
     const splitArguments = splitArgumentsString(argumentsString) || []
     const key = splitArguments[0].replaceAll("\"", "")
 
-    const eachLanguageStrings: string[] = []
-    get(selectedLanguages).forEach((language) => {
+    const genLanguageString = (language: Language): string => {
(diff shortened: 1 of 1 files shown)


app/javascript/src/utils/compiler/translations.ts
```diff
@@ -24,16 +24,15 @@ export function convertTranslations(joinedItems: string, singleLanguageOverride:
     const splitArguments = splitArgumentsString(argumentsString) || []
     const key = splitArguments[0].replaceAll("\"", "")
 
-    const eachLanguageStrings: string[] = []
-    get(selectedLanguages).forEach((language) => {
+    const genLanguageString = (language: Language): string => {
```
Mitcheljager, line 27: No abbreviations >:(

Mitcheljager (approved):
Good find, but please don't use abbreviations!

app/javascript/src/utils/compiler/translations.ts
netux: Ok u.u

Mitcheljager (approved):

PR #520, opened 2025-03-11
fix(editor code actions): make transformParameterObjectsIntoPositionalParameters code action appear again

netux:
- Search for the correct completion type when calculating for which completion we should provide this code action
- Adjust z-index of lightbulb button to be +1 the z-index of the CodeMirror gutter


Mitcheljager (approved):
👍

PR #521, opened 2025-03-11
feature(editor linter): warn about unknown translation keys, allow for creating translation key in-place

netux:
Also:
- Refactor `checkTranslations()` to allow for different diagnostics `from`, `to`, and `severity`
- When opening a modal, pass any properties under `$modal.parameters` as parameters for the modal component
  - Used here to open the translation keys modal with the new key already selected

### DIFF
--- a/app/javascript/src/components/editor/Modals/Modals.svelte
+++ b/app/javascript/src/components/editor/Modals/Modals.svelte
@@ -21,5 +21,5 @@
 </script>
 
 {#if component}
-  <svelte:component this={component} {... $modal.parameters ?? {}} />
+  <svelte:component this={component} {...$modal.parameters || {}} />
 {/if}
--- a/app/javascript/src/components/editor/Modals/Modals.svelte
+++ b/app/javascript/src/components/editor/Modals/Modals.svelte
@@ -21,5 +21,5 @@
 </script>
 
 {#if component}
-  <svelte:component this={component} {...$modal.parameters || {}} />
+  <svelte:component this={component} {...$modal.props || {}} />
 {/if}
--- a/app/javascript/src/components/editor/Modals/Modals.svelte
+++ b/app/javascript/src/components/editor/Modals/Modals.svelte
@@ -21,5 +21,5 @@
 </script>
 
 {#if component}
-  <svelte:component this={component} />
+  <svelte:component this={component} {... $modal.parameters ?? {}} />
 {/if}
--- a/app/javascript/src/lib/OWLanguageLinter.ts
+++ b/app/javascript/src/lib/OWLanguageLinter.ts
@@ -3,7 +3,7 @@ import { completionsMap, currentItem, modal, subroutinesMap, workshopConstants }
 import { get } from "svelte/store"
 import { getFirstParameterObject } from "@utils/compiler/parameterObjects"
 import type { EditorView } from "codemirror"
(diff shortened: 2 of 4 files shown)


netux:
Code is ugly. I'm open to suggestions about refactoring (mainly moving things around) :D

app/javascript/src/components/editor/Modals/Modals.svelte
```diff
@@ -21,5 +21,5 @@
 </script>
 
 {#if component}
-  <svelte:component this={component} />
+  <svelte:component this={component} {... $modal.parameters ?? {}} />
```
Mitcheljager, line 24: 2 things:
- Spread has a space (I didn't even know that was allowed)
- I think it would make more sense to use `||` rather than `??`. If the value is falsey, it will error.
```suggestion
  <svelte:component this={component} {...$modal.parameters || {}} />
```

app/javascript/src/lib/OWLanguageLinter.ts
```diff
@@ -1,11 +1,13 @@
 import { findRangesOfStrings, getClosingBracket, getPhraseFromPosition, matchAllOutsideRanges, splitArgumentsString } from "@utils/parse"
-import { completionsMap, subroutinesMap, workshopConstants } from "@stores/editor"
+import { completionsMap, currentItem, modal, subroutinesMap, workshopConstants } from "@stores/editor"
 import { get } from "svelte/store"
 import { getFirstParameterObject } from "@utils/compiler/parameterObjects"
 import type { EditorView } from "codemirror"
-import type { Diagnostic } from "@codemirror/lint"
-import type { Severity } from "@src/types/editor"
+import { type Diagnostic } from "@codemirror/lint"
```
Mitcheljager, line 6: Any reason for changing this? The others are all `import type`

Mitcheljager (changes requested):

app/javascript/src/lib/OWLanguageLinter.ts
```diff
@@ -242,40 +244,106 @@ function checkMixins(content: string): void {
...
+              newTranslation[currentDefaultLanguage] = translateKey
+            }
+
+            translationKeys.update((keys) => ({
+              ... keys,
+              [translateKey]: newTranslation
+            }))
+
+            modal.show(Modal.TranslationKeys, {
+              parameters: {
```
Mitcheljager, line 333: I think it might more sense to call this `props`, as that is what the Svelte component parameters are called

app/javascript/src/lib/OWLanguageLinter.ts
netux: That's my IDE doing IDE stuff. I think its because I had prettier setup.

app/javascript/src/lib/OWLanguageLinter.ts
netux: _An abbreviation?? In this code base?!_

---

Ah, I didn't remember how Svelte called them, and I was too lazy to look it up.
Sure, I'll change it.

app/javascript/src/lib/OWLanguageLinter.ts
netux:
```suggestion
import type { Diagnostic } from "@codemirror/lint"
```

Mitcheljager (approved):

PR #522, opened 2025-03-11
fix(editor linter): don't extract arguments from @translate() one character too short

netux:
Fixes erroneous diagnostics for the @translate()'s key argument type when @translate() takes no additional arguments


Mitcheljager (approved):
Sneaky

PR #523, opened 2025-03-11
 feat(translations): Add static modifier to @translate 

Mitcheljager:
@netux 

This adds a modifier for the `@translate` feature, allowing you to output a string directly, rather than `Custom String`. This is handy when compiling to a single language and using `@translate` in places it normally wouldn't work. One example of that is the description in settings.

Static keys can not have arguments, as that makes no sense. A linter was added for this.

image
image
image
image

### DIFF
--- a/app/javascript/src/lib/OWLanguageLinter.ts
+++ b/app/javascript/src/lib/OWLanguageLinter.ts
@@ -317,6 +317,15 @@ function checkTranslations(content: string): void {
       })
       continue
     }
+    if (translateArguments.length > 1 && isStatic) {
+      diagnostics.push({
+        from: match.index + match[0].length + translateArguments[0].length,
+        to: translateEndParenthesisIndex,
+        severity: "warning",
+        message: "Additional arguments in static translations will have no effect."
+      })
+      continue
+    }
 
     // Provide option to create translation key via linter
     const translateKey = translateArguments[0].substring(1, translateArguments[0].length - 1)
--- a/app/javascript/src/lib/OWLanguageLinter.ts
+++ b/app/javascript/src/lib/OWLanguageLinter.ts
@@ -315,7 +315,7 @@ function checkTranslations(content: string): void {
     }
 
     if (translateArguments.length > 1 && isStatic) {
-      const secondArgumentStart = translateArgumentsString.indexOf(translateArguments[1])
+      const secondArgumentStart = translateArgumentsString.indexOf(translateArguments[1], translateArguments[0].length)
 
       diagnostics.push({
         from: translateStartParenthesisIndex + secondArgumentStart + 1,
--- a/app/javascript/src/lib/OWLanguageLinter.ts
+++ b/app/javascript/src/lib/OWLanguageLinter.ts
@@ -244,10 +244,11 @@ function checkMixins(content: string): void {
 }
 
 function checkTranslations(content: string): void {
-  const regex = /(@translate)(\.static)?\s*\(/g
+  const regex = /(@translate(?<isStatic>\.static)?)\s*\(/g
+
   let match
   while ((match = regex.exec(content)) != null) {
-    const isStatic = !!match[2] // If `.static` is given
+    const isStatic = !!match.groups?.isStatic // If `.static` is given
 
     let walk = match.index
     let parenthesisBeforeAtIndex = -1
--- a/app/javascript/src/lib/OWLanguageLinter.ts
+++ b/app/javascript/src/lib/OWLanguageLinter.ts
@@ -315,8 +315,10 @@ function checkTranslations(content: string): void {
     }
 
     if (translateArguments.length > 1 && isStatic) {
+      const secondArgumentStart = translateArgumentsString.indexOf(translateArguments[1])
+
       diagnostics.push({
-        from: match.index + match[0].length + translateArguments[0].length,
+        from: translateStartParenthesisIndex + secondArgumentStart + 1,
         to: translateEndParenthesisIndex,
         severity: "warning",
         message: "Additional arguments in static translations will have no effect."
--- a/app/javascript/src/lib/OWLanguageLinter.ts
+++ b/app/javascript/src/lib/OWLanguageLinter.ts
@@ -250,44 +250,49 @@ function checkTranslations(content: string): void {
   while ((match = regex.exec(content)) != null) {
     const isStatic = !!match.groups?.isStatic // If `.static` is given
 
-    let walk = match.index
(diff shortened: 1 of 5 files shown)


app/javascript/src/lib/OWLanguageLinter.ts
```diff
@@ -244,9 +244,11 @@ function checkMixins(content: string): void {
 }
 
 function checkTranslations(content: string): void {
-  const regex = /(@translate)[\s\n]*\(/g
+  const regex = /(@translate)(\.static)?\s*\(/g
```
netux, line 247: Match group 1 is used for knowing the length of the whole `@translate`/`@translate.static` directive (without the parenthesis or any whitespace after it). This is used in the diagnostics `to` field, for example.

```suggestion
  const regex = /(@translate(\.static)?)\s*\(/g
```

app/javascript/src/lib/OWLanguageLinter.ts
```diff
@@ -244,9 +244,11 @@ function checkMixins(content: string): void {
...
   let match
   while ((match = regex.exec(content)) != null) {
+    const isStatic = !!match[2] // If `.static` is given
```
netux, line 250: Could use named groups instead

```ts
const regex = /...(?<isStatic>\.static).../
const match = regex.exec(content)
const isStatic = !!match.groups.isStatic
```

app/javascript/src/lib/OWLanguageLinter.ts
```diff
@@ -261,7 +263,7 @@ function checkTranslations(content: string): void {
       walk--
     }
 
-    if (parenthesisBeforeAtIndex === -1) {
+    if (!isStatic && parenthesisBeforeAtIndex === -1) {
```
netux, line 266: We could skip a lot of work here by wrapping the calculation of `parenthesisBeforeAtIndex` and this in an `if (!isStatic)` too

app/javascript/src/lib/OWLanguageLinter.ts
```diff
@@ -277,7 +279,7 @@ function checkTranslations(content: string): void {
     const phrase = getPhraseFromPosition(line, parenthesisBeforeAtIndex - 1)
     const acceptedPhrases = ["Create HUD Text", "Create In-World Text", "Create Progress Bar HUD Text", "Create Progress Bar In-World Text", "Set Objective Description", "Big Message", "Small Message"]
     if (phrase?.text.includes("include")) continue
-    if (phrase?.text && !acceptedPhrases.includes(phrase.text)) {
+    if (!isStatic && phrase?.text && !acceptedPhrases.includes(phrase.text)) {
```
netux, line 282: Ditto. You can calculate the phrases at position only when `isStatic === false`

app/javascript/src/lib/OWLanguageLinter.ts
```diff
@@ -295,6 +297,16 @@ function checkTranslations(content: string): void {
     const translateArguments = splitArgumentsString(content.substring(translateStartParenthesisIndex + 1, translateEndParenthesisIndex - 1))
     if (translateArguments.length === 0) continue
 
+    if (translateArguments.length > 1 && isStatic) {
```
netux, line 300: I'd say this should go after checking the type of `translateArguments[0]`

app/javascript/src/lib/OWLanguageLinter.ts
```diff
@@ -295,6 +297,16 @@ function checkTranslations(content: string): void {
...
+      diagnostics.push({
+        from: match.index + match[0].length + translateArguments[0].length,
```
netux, line 302: Could the diagnostic start from the second argument, instead of the comma after the first argument?
I can imagine that's a little harder to do since the argument splitting is handled by `splitArgumentsString`, but maybe you can think of some quick solution?

diagnostic highlighting from the comma after the first argument, very ugly

app/javascript/src/utils/compiler/compile.ts
```diff
@@ -28,10 +28,9 @@ export function compile(overrideContent = "", singleLanguageOverride = null): st
   joinedItems = evaluateForLoops(joinedItems)
   joinedItems = evaluateEachLoops(joinedItems)
   joinedItems = evaluateConditionals(joinedItems)
-  joinedItems = convertTranslations(joinedItems, singleLanguageOverride)
 
   const variables = compileVariables(joinedItems)
   const subroutines = compileSubroutines(joinedItems)
 
-  return settings + variables + subroutines + joinedItems
+  return convertTranslations(settings + variables + subroutines + joinedItems, singleLanguageOverride)
```
netux, line 35: Really only `settings` needs its translations converted. Maybe we can just call `convertTranslations()` in `joinedItems` and in `settings`? It'd technically be less work.

```ts
settings = convertTranslations(settings, singleLanguageOverride)

joinedItems = convertTranslations(joinedItems, singleLanguageOverride)

// ...

return settings + variables + subroutines + joinedItems
```

app/javascript/src/utils/compiler/translations.ts
```diff
@@ -8,9 +8,50 @@ export function convertTranslations(joinedItems: string, singleLanguageOverride:
...
+  }
+
+  return joinedItems
+}
+
+/**
+ * Replace @translate("Key") with Custom String("Value"), either compiled to a single language or a with dynamic languages.
+ */
+function replaceDynamicTranslationKeys(joinedItems: string, singleLanguageOverride: Language | null): string {
   const regex = /@translate/g
```
netux, line 52: Just making sure :eyes:

```suggestion
  const regex = /@translate(?!\.static)/g
```

netux (changes requested):

app/javascript/src/utils/compiler/compile.ts
Mitcheljager: Yes, but I'm choosing readability over speed here. Performance is not an concern when compiling.

app/javascript/src/lib/OWLanguageLinter.ts
```diff
@@ -315,8 +315,10 @@ function checkTranslations(content: string): void {
     }
 
     if (translateArguments.length > 1 && isStatic) {
+      const secondArgumentStart = translateArgumentsString.indexOf(translateArguments[1])
```
netux, line 318: Smart! One little suggestion though

```suggestion
      const secondArgumentStart = translateArgumentsString.indexOf(translateArguments[1], translateArguments[0].length)
```

Make the search start from the end of the first argument. Just in case the first argument is the same as the second one or something lol.

PR #524, opened 2025-03-12
feature(editor translations): add import and export functionality

netux:
- Parse/unparse using papaparse
- Add a bunch feedback to the user, making use of the data papaparse provides
- Import as either CSV or TSV; export as TSV (pastable into Excel)

Also:
- Add types to \<TranslationKeysModal> (because I like working with types) and \<Model> (needed by \<TranslationKeysModal>)
- Adjust styling of alerts so they can have new lines in time (used prominently in the import function)


netux:
I want try adding import and export to/from a file. But that'll probably be a thing for tomorrow me to work at.

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
```diff
@@ -28,6 +32,109 @@
...
     newKeyInput.value = ""
   }
+
+  const DELIMITER_TO_FILE_TYPE = new Map([
+    [",", "CSV"],
+    ["\t", "TSV"],
+    [null, "TSV/CSV"]
+  ])
+
+  async function importKeysFromClipboard() {
```
Mitcheljager, line 42: Functions should have a return type (There should be a linter for this, but it's possible the linter doesn't work in Svelte files)

```suggestion
  async function importKeysFromClipboard(): void {
```

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
```diff
@@ -28,6 +32,109 @@
...
+      const { key, ... translations } = entry
+      newKeys[key] = translations
+
+      return newKeys
+    }, {} as Record<string, TranslationKey>)
+
+    addAlert("Translation keys successfully imported from clipboard!")
+  }
+
+  async function exportKeys() {
```
Mitcheljager, line 123: Same for this one
```suggestion
  async function exportKeys(): void {
```

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
```diff
@@ -76,6 +183,11 @@
         {/if}
 
         <button on:click={addKey} class="button button--small button--square w-100 mt-1/8">Create</button>
+
+        <div class="button-group mt-1/4">
+          <button on:click={() => importKeysFromClipboard()} class="button button--small button--secondary button--square w-100" data-tooltip="From CSV/TSV in your clipboard">Import</button>
+          <button on:click={() => exportKeys()} class="button button--small button--secondary button--square w-100" data-tooltip="To TSV (for Excel)">Export</button>
+        </div>
```
Mitcheljager, line 190: To clean thing up a bit, I think it makes more sense to move this and the new function above to a new Svelte component.

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
Mitcheljager, line 42: Since this is quite a destructive function, it might be good to remind the user that they are about to override their current keys.

```js
if (!confirm("Importing will override your current translations keys")) return
```

Mitcheljager (changes requested):
Looks good overall, seems to work well. I had some minor formatting complaints. Perhaps most important is moving everything to a separate component to keep things relatively compartmentalized

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
```diff
@@ -1,17 +1,21 @@
-<script>
+<script lang="ts">
   import Modal from "@components/editor/Modals/Modal.svelte"
   import TranslationKeysEditStrings from "@components/editor/TranslationKeys/TranslationKeysEditStrings.svelte"
   import TranslationKeysSelectLanguages from "@components/editor/TranslationKeys/TranslationKeysSelectLanguages.svelte"
   import { translationKeys, orderedTranslationKeys, selectedLanguages } from "@stores/translationKeys"
   import { copyValueToClipboard } from "@src/copy"
   import { submittable } from "@components/actions/submittable"
+  import { addAlert } from "@src/lib/alerts"
+  import papaparse from "papaparse"
```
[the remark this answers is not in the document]
netux: Good idea!

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
netux: Agreed!

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
netux: I was thinking since the keys aren't fully overwritten until the user saves, then there is always a way back.

Also, I don't really like the idea of the browser dialog leading directly to copying to clipboard (especially since the button is just labeled "Confirm").
Maybe if it were its own special in-app modal with a button like "I get it, confirm anyways"?

app/javascript/scss/elements/_dropdown.scss
```diff
@@ -51,6 +51,11 @@
     right: auto;
   }
 
+  &--center {
+    right: 50%;
+    translate: 50% 0;
```
Mitcheljager, line 56: `translate` isn't well enough supported (we supposed Safari 13+)

```suggestion
    transform: translateX(50%);
```

Mitcheljager (changes requested):
Good stuff! I did have 1 minor thing

PR #526, opened 2025-03-14
fix(editor linter): use forceLinting() instead of $currentItem.forceUpdate to rerun linter inside diagnostic action callback, patch lint plugin to support this

netux:
`@codemirror/lint` has a (bug?)/annoying behavior where `forceLinting()` would do nothing if it is run ([by default](https://github.com/codemirror/lint/blame/2ffcff1ed78c13faedbe40e7845cc384c3ddd26e/src/lint.ts#L363)) within 750ms of linting being done (and also within 750ms of running diagnostic action too, it seems).

This made me consider setting `$currentItem.forceUpdate` to force a full update on the editor.
I thought it was overkill, but it fixed my issue.

Turns out calling setting `forceUpdate` is now causing other issues, like the editor's content going back to its initial text.

This PR switches that for the original plan of using `forceLinting()`, while also patching the issue there (thanks for having patch-package!)


Mitcheljager (approved):
Makes sense! Let's hope CodeMirror didn't have a real good reason for it being the way that it is.

PR #527, opened 2025-03-14
feat(editor translations modal): miscellaneous improvements to the modal

netux:
Trying to make this modal a bit easier to use :)

See commits for the individual changes

> [!NOTE]
> Key creation flow I teased in Discord is in a separate PR: #528


app/javascript/src/components/editor/TranslationKeys/TranslationKeysEditStrings.svelte
```diff
@@ -32,9 +33,9 @@
     $translationKeys = { ...$translationKeys }
 
     dispatch("updateKey", value)
-  }
+  }, 300)
 
-  function removeKey() {
+  const removeKey = () => {
```
Mitcheljager, line 38: Regular functions are fine. Why did you choose to refactor this to be an arrow function?

app/javascript/scss/editor/_translation-settings.scss
```diff
@@ -13,11 +17,24 @@
   background: $bg-darker;
 }
 
+.translation-settings__keys-container {
+  height: 100%;
+  overflow-y: auto;
+  margin-inline: -1rem;
+  padding-inline: 1rem;
+
```
Mitcheljager, line 25: I'd prefer this use regular margins and padding, it's not like we care about different writing modes.

```suggestion
  margin: 0 -1rem;
  padding: 0 1rem;

```

Mitcheljager (changes requested):
Looks good overall, but there is a design issue. On small screens the key container element becomes impossible to scroll:
image

app/javascript/src/components/editor/TranslationKeys/TranslationKeysEditStrings.svelte
netux: I was forced to by `renameKey`, since I wrapped it inside a `debounce`. Admittedly, I could've left that as an anonymous function as well, but a bit of personal preference swooped in as well.

I checked on a random file that we also sometimes use arrow functions at the root level of Svelte component scripts, just to be sure you'd maybe be okay with it 😅.

Since `renameKey` was changed, it didn't seem nice to have `removeKey` be a regular function too, so I changed that as well.

In short: personal preference + a change that somewhat justifies it.

app/javascript/scss/editor/_translation-settings.scss
netux: Those are not truly equal though: `margin/padding-inline` only changes the margin/padding on the (in english) horizontal plane, while yours also sets them to zero on the vertical plane.

I didn't care about writing modes either, I was just using the shorthand that would avoid repeating the margin and padding for both left and right, while also avoiding overwriting the margin/padding in other axis.

But I'm fine with your suggestion regardless.
I honestly thought you'd point out the negative margin 😆. But I take you know why I did that? Maybe I should add a comment.

Mitcheljager: Yep, that's totally fine, to counteract the wrapper padding

netux: To clarify: you still prefer `margin/padding: 0 n` over `margin-inline: n`?

app/javascript/scss/editor/_translation-settings.scss
```diff
@@ -20,8 +20,12 @@
...
   height: 100%;
   overflow-y: auto;
-  margin-inline: -1rem;
-  padding-inline: 1rem;
+
+  // Counteract padding from .translation-settings__aside to move the scrollbar
+  // to the edge of it, then reapply the padding so keys are not touching the
+  // edge of the .translation-settings__aside also.
+  margin-inline: -($margin * 0.25);
+  padding-inline: ($margin * 0.25);
```
Mitcheljager, line 28: Still though, let's not use -inline.

Mitcheljager (approved):
Good stuff, thanks!

PR #528, opened 2025-03-14
feat(editor translations modal): implement new key creation flow

netux:
❗ Includes #527. Please merge that first!

Replace the old inline form below keys with an "Add" button that sets up \<TranslationKeysEditStrings> in a "create key" mode.

While in this mode, the key input is empty, and prompts the user to write the new of the new key.

translations modal with the <edit strings> component in "create key" mode

As soon as a new key is submitted, the key is created and translation \<textarea>s will appear.

translations modal after submitting a new key, showing the new key being created also in the keys list on the side


app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
```diff
@@ -44,9 +27,12 @@
         Select languages ({$selectedLanguages.length})
       </button>
 
-      {#if Object.keys($orderedTranslationKeys).length}
-        <h4 class="mb-1/8"><strong>Keys</strong></h4>
+      <h4 class="flex align-center justify-between mb-1/8">
+        <strong>Keys</strong>
+        <button class="button button--primary button--small button--square" on:click={() => { selectedKey = createNewTranslationKey; showLanguageSettings = false }}>Add</button>
```
Mitcheljager, line 31: Calling creaetNewTranslationKey like that here will assign the function, not the value, right?
```suggestion
        <button class="button button--primary button--small button--square" on:click={() => { selectedKey = createNewTranslationKey(); showLanguageSettings = false }}>Add</button>
```

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
```diff
@@ -60,27 +46,11 @@
             </div>
           {/each}
         </div>
+      {:else}
+        <p class="m-0">No keys</p>
```
Mitcheljager, line 49: In Svelte you can also add an `else` after the `each`, which will be true if the array has no length

(Github is being annoying, can't give a proper suggestion)

```svelte
<div class="translation-settings__keys-container">
  {#each Object.keys($orderedTranslationKeys) as key}
    <div>
       ...
    </div>
  {:else}
     <p class="m-0">No keys</p>
  {/each}
</div>
```

Mitcheljager, line 49: Perhaps a slightly friendlier empty state would be nice

```suggestion
        <p class="m-0">You have not yet created any keys.</p>
```

Or something like that

app/javascript/src/components/editor/TranslationKeys/TranslationKeysEditStrings.svelte
```diff
@@ -4,20 +4,32 @@
...
 
-  const renameKey = debounce(() => {
+  $: {
+    createNewKeyMode = selectedKey === createNewTranslationKey
+
+    selectedKeyNameInExamples = createNewKeyMode ? "…" : selectedKey
+
+    if (createNewKeyMode && createOrRenameInput)
+      createOrRenameInput.focus()
+  }
```
Mitcheljager, line 25: I'm personally not the biggest fan of big reactive statements like this. I prefer to keep it to one line if possible and extract the rest to a function, that way you have better control over what is and isn't reactive

app/javascript/src/components/editor/TranslationKeys/TranslationKeysEditStrings.svelte
```diff
@@ -47,19 +67,26 @@
 
 <div class="sticky top-0 well well--dark block p-1/4 mb-1/4">
   <div class="flex gap-1/4 align-center">
-    <input class="form-input" value={selectedKey} bind:this={renameInput} on:input={renameKey} />
+    <input
+      class="form-input"
+      value={createNewKeyMode ? "" : selectedKey}
+      bind:this={createOrRenameInput}
+      on:input={!createNewKeyMode && debounce(createOrRenameKey, 300)}
+      on:change={createNewKeyMode && createOrRenameKey}/>
```
Mitcheljager, line 75: Does it need to be on both input and change? That seems odd

app/javascript/src/lib/translations.ts
```diff
@@ -0,0 +1 @@
+export const createNewTranslationKey = Symbol("create new translations key")
```
Mitcheljager, line 1: Why did you choose a symbol here?

app/javascript/src/components/editor/TranslationKeys/TranslationKeysEditStrings.svelte
```diff
@@ -47,19 +67,26 @@
...
     <button class="button button--danger button--small button--square" on:click={removeKey}>Remove</button>
   </div>
 
-  {#if error}
+  {#if createNewKeyMode}
+    <div class="text-orange mt-1/8">Give your translation a key</div>
```
Mitcheljager, line 80: Perhaps a `transition:slide={{ duration: 100 }}` would look nice here, less jumpy

Mitcheljager (changes requested):
Looks like a nice improvement overall, but I did have a few points (as is tradition)

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
netux: `createNewTranslationKey` is a symbol defined defined in `utils/translations.ts`

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
netux: I was going for less verbosity, but sure :+1:

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
Mitcheljager: Ah of course, it sounds like a function, like it creates something

app/javascript/src/components/editor/TranslationKeys/TranslationKeysEditStrings.svelte
netux: I was wondering if I should maybe add a comment here. I think for UX this is pretty nice:

- When creating a new key, the key won't be created until the user has pressed the Enter key (or moved focus away from the input)
- But when editing an already existing key, the explicit submission shouldn't be required (like with the translation \<textarea>s)

---

This is nice for users, but I agree its a bit confusing for developers. The main issue is I'm reusing the same input here for both. Maybe I should extract the input out, and make the different modes more separate in the code?

app/javascript/src/lib/translations.ts
netux: A nullish `selectedKey` means "no key selected". A `selectedKey` that is a string means the key of that name if selected. There were no other (sensible) values left, which is where Symbols come in (I think).

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
netux: Maybe I should drop the `create`, and add `Symbol` at the end?: `newTranslationKeySymbol`

app/javascript/src/components/editor/TranslationKeys/TranslationKeysEditStrings.svelte
Mitcheljager: Got it, that does make sense. Maybe there would be a way to do it both on input (or maybe keydown?) but extracted to a function somehow? Otherwise leaving a comment will do

app/javascript/src/components/editor/Modals/TranslationKeysModal.svelte
Mitcheljager: Sure! Just `newTranslationKey` works too I think

app/javascript/src/components/editor/TranslationKeys/TranslationKeysEditStrings.svelte
netux: How about this? 4bf02b78c4011e1d5f3bc7c5d304ff3ba05df249

Mitcheljager: Nice, that's much more readable

app/javascript/src/components/editor/TranslationKeys/TranslationKeysEditStrings.svelte
netux: Like so? Or just to the "Give your translation a key"?

new key instruction and error slide animation

Mitcheljager: Yeah that's nice!

netux:
I changed the view a bit: now the usage instructions are hidden, and the create key instructions are where the usage instructions would've been

what I described above, lol

Mitcheljager (approved):
Looks nice!

PR #531, opened 2025-03-24
feat(editor linter): add an "Array" as accepted Phrase where a @translate directive can be

netux:
See https://discord.com/channels/570672959799164958/574537911114924043/1353552205839597589


CactusPuppy:
@Mitcheljager Is there any reason we never merged this?

Mitcheljager:
@CactusPuppy Following the Discord discussion, it's because it doesn't actually work. I don't remember at all though.

PR #532, opened 2025-04-04
fix(editor): Remove comments before extracting variables

Mitcheljager:
Variables would still be evaluated from comments. This commit fixes that by removing comments before extracting variables.

Before | After
--- | ---
image | image


netux (approved):
lgtm ezClap

PR #538, opened 2025-07-24
fix(editor compiler): do not try to parse operators in expressions wrapped in "/"s

netux:
Fixes an issue where in an @if condition, using a match operator with the righthand-side being a RegExp with an exclamation mark ("!"), the exclamation mark would be considered as an extra operator and cause issues in the evaluation.


Mitcheljager (approved):
Nice!

